PR #4, opened 2017-06-01
First step: Use Kinto without encrypting on each keyStroke.

Natim:


src/background.js
```diff
@@ -1,6 +1,6 @@
 const REDIRECT_URL = browser.identity.getRedirectURL();
 const CLIENT_ID = 'c6d74070a481bc10';
-const SCOPES = ['profile keys'];
+const SCOPES = ['kinto profile keys'];
```
vladikoff: I think what we are aiming for here with the encryption doc is: 

`const SCOPES = ['notes profile'];` or 
`const SCOPES = ['app:notes profile'];` or 
`const SCOPES = ['app:notes:write profile'];`

the FxA side of things would not know anything about 'kinto' scopes

Natim: The scope just need to be in sync with the kinto server configuration because it will verify that the scope is correct. For kinto.dev the mandatory scope is kinto :)

vladikoff: ah, yea that should work 👍

src/sidebar/panel.js
```diff
@@ -6,15 +10,73 @@ var quill = new Quill('#editor', {
...
+        {"attributes": {"size": "large"},
+         "insert": "Sync them to our Android app: http://mzl.la/notes"},
+        {"attributes": {"list": "ordered"}, "insert": "\n"}]});
+  } else {
+    quill.setContents(data["notes"]);
+  }
+}
+
+browser.storage.local.get(["bearer", "keys", "notes"], function(data) {
+  // If we have a bearer, we try to save the content.
```
vladikoff: for next steps, do we need to check if the user has the 'default' placeholder text, e.g `Welcome!` and overwrite it after a login? I guess if it is not default text then we merge the records on login?

vladikoff:
Good start @Natim !!

PR #32, opened 2017-06-13
Add prettier for code beautification

### LINKED ISSUE #10: Add prettier to auto-format code?
Here was my local config, although the indenting is a bit different between ESLint and Prettier, so I think we'd need to disable the indenting rules in ESLint.

```js
"format": "prettier 'src/**/!(kinto-http|quill|quill.snow).{css,js}' --single-quote --write",
```

And here's my output (showing the processed files):

```sh
➜  notes git:(fix-eslint) $ npm run format

> notes@1.0.0 format /Users/pdehaan/dev/github/mozilla/notes
> prettier 'src/**/!(kinto-http|quill|quill.snow).{css,js}' --single-quote --write

src/background.js 258ms
src/sidebar/options/options.js 23ms
src/sidebar/panel.js 190ms
src/sidebar/styles.css 120ms
```

pdehaan:
Fixes #10

### DIFF
--- a/src/sidebar/styles.css
+++ b/src/sidebar/styles.css
@@ -1,23 +1,23 @@
 html {
-	height: 100%;
+  height: 100%;
 }
 
 body {
-	height: 100%;
-	margin: 0;
-	overflow: hidden;
-	padding: 0;
+  height: 100%;
+  margin: 0;
+  overflow: hidden;
+  padding: 0;
 }
 
 div#editor {
-	height: 90%;
-	height: calc(100% - 70px);
+  height: 90%;
+  height: calc(100% - 70px);
 }
 
 button#enableSync {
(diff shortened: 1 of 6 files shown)


vladikoff:
Needs rebase?

src/sidebar/styles.css
```diff
@@ -1,23 +1,23 @@
...
+  margin: 0;
+  overflow: hidden;
+  padding: 0;
 }
 
 div#editor {
-	height: 90%;
-	height: calc(100% - 70px);
+  height: 90%;
+  height: calc(100% - 70px);
```
pdehaan: Just noticed that we define `height` twice here... I'll do some CSS linting and file a separate bug.

Issue #17, opened 2017-06-12
Add Circle-CI integration 


pdehaan:
We should run ESLint and run some tests on each commit to ensure we keep it all kosher.

PR #50, opened 2017-06-15
Add CircleCI

vladikoff:
Fixes https://github.com/mozilla/notes/issues/17

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,7 @@
+machine:
+  node:
+    version: 6
+
+test:
+  override:
+    - echo "Tests WIP"
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Firefox Notes
 
+[![CircleCI](https://circleci.com/gh/mozilla/notes.svg?style=svg)](https://circleci.com/gh/mozilla/notes)
+
 ## What it does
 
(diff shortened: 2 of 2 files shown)


circle.yml
```diff
@@ -0,0 +1,7 @@
+machine:
+  node:
+    version: 6
+
+test:
+  override:
+    - echo "Tests WIP"
```
pdehaan: Add `- npm run lint`?

pdehaan:
Not sure if this is extreme overkill, but over in [**SnoozeTabs** land](https://github.com/bwinton/SnoozeTabs/blob/master/circle.yml), I believe we do a build of the add-on in Circle and then copy the generated XPI into the `$CIRCLE_ARTIFACTS` directory. That way we can grab old add-ons on a per-run basis by grabbing them from Circle (see https://circleci.com/gh/bwinton/SnoozeTabs/560#artifacts/containers/0).

Again, not sure if that is needed, considering how trivial it'd be to do a git checkout of a previous commit and then build it locally, but this may be useful for contributors or whatever.

vladikoff:
@pdehaan that would be nice!

circle.yml
vladikoff: @pdehaan that exploded with errors, so we gonna have to add that later

PR #70, opened 2017-06-16
Use JWE format to store the encrypted content in Kinto.

### LINKED ISSUE #46: Use JWE for notes encryption
[JSON Web Encryption](https://tools.ietf.org/html/rfc7516) is an industry standard to stored blobs of data in encrypted form inside a JSON envelope. We should consider using it instead of a custom encryption format for notes.

Natim:
Fixes #46 

I didn't noticed that master changed that much overnight so I had a really hard time rebasing this changes.


src/sidebar/panel.js
```diff
@@ -31,27 +51,34 @@ function base64ToArrayBuffer(base64) {
...
   });
+
+  const ciphertextBuf = new Uint8Array(data.length);
   data.forEach((byte, i) => {
-    buf[ivLen + i] = byte;
+    ciphertextBuf[i] = byte;
+  });
+
+  return JSON.stringify({
+    protected: window.btoa(JSON.stringify({ enc: 'A256GCM' })),
```
vladikoff: from the spec: `  "protected", with the value BASE64URL(UTF8(JWE Protected Header))`, so I think we need to use some string `...ToBase64URL` instead of just `btoa` here?

Natim: Oh good catch, we have a function for that already: 
https://github.com/mozilla/notes/pull/70/files#diff-1abf0f3e5a69394934ea8f7b5c521805R24

vladikoff:
@Natim I'm getting ghost text spasms with this: 

login

:shaky:  :)

vladikoff:
Follow up issue: https://github.com/mozilla/notes/issues/73

PR #77, opened 2017-06-16
Use JS-JOSE instead of custom crypto code.

### LINKED ISSUE #75: Switch custom crypto to use js-jose
See https://github.com/square/js-jose/blob/master/examples/jose-jwe-example2.html

It will greatly simplfiy the code of the add-ons and we would benefit from a shared security monitoring of the lib.

Natim:
Fixes #75

### DIFF
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -6,3 +6,4 @@ cp node_modules/testpilot-metrics/testpilot-metrics.js src/vendor
 cp node_modules/kinto-http/dist/kinto-http.js src/sidebar/vendor
 cp node_modules/quill/dist/quill.js src/sidebar/vendor
 cp node_modules/quill/dist/quill.snow.css src/sidebar/vendor
+cp node_modules/jose-jwe-jws/dist/jose.js src/sidebar/vendor
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -9,6 +9,9 @@ extends:
 globals:
   KintoClient: false
   Quill: false
+  Jose: false
+  JoseJWE: false
+  Metrics: false
 
 root: true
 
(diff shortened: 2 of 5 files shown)


vladikoff:
@Natim nice!!

scripts/postinstall.sh
```diff
@@ -6,3 +6,4 @@ cp node_modules/testpilot-metrics/testpilot-metrics.js src/vendor
 cp node_modules/kinto-http/dist/kinto-http.js src/sidebar/vendor
 cp node_modules/quill/dist/quill.js src/sidebar/vendor
 cp node_modules/quill/dist/quill.snow.css src/sidebar/vendor
+cp node_modules/jose-jwe-jws/dist/jose.js src/sidebar/vendor
```
pdehaan: Do you want _jose.js (51K)_ or the _jose.**min**.js (19K)_?

```sh
➜  notes git:(master) ✗ ls -lash node_modules/jose-jwe-jws/dist

104 -rw-r--r--   1 pdehaan  staff    51K May 20 10:48 jose-commonjs.js
104 -rw-r--r--   1 pdehaan  staff    51K May 20 10:48 jose-testing.js
104 -rw-r--r--   1 pdehaan  staff    51K May 20 10:48 jose.js
 40 -rw-r--r--   1 pdehaan  staff    19K Oct 25  2016 jose.min.js
```

jvehent:
Less crypto code == win!
One thing I like to do when writing crypto code is to perform a decryption with a separate library that implements the same standard. For example, you could use PyJWE to decrypt a document encrypted with js-jose. That would help verify compatibility.

scripts/postinstall.sh
Natim: I asked myself the question but since we didn't use the minified version for the other ones I didn't. Also the file is served locally so I don't think 30K are that important is it?

Issue #73, opened 2017-06-16
Handle If-Match headers to avoid overridding when writing to Kinto


Natim:
Now that we are polling changes from Kinto less often, it could happen that we are storing an update in Kinto without having loaded the last version from it.

To avoid that we should keep track of the Etag from the last time we read Kinto to avoid overriding if an update occurred in between.

The current implementation only handles conflict on loading from Kinto (with regards to the local pad) but not when writing to it.

Natim:
After talking about it this morning, we want to handle multiple cases here:

- Raise conditions, when to pad where edited and try to sync at the same time. (Merge conflict)
- When we synced the content of the pad but the key was reset in between, which means what we had synced in Kinto before the password reset in FxA was not merged (because unreadable with the new keys) In that case we want to detect a merge conflict with the local content.

This means we need to store the kid as well as the etag locally.

Natim:
Fixed in #241

PR #85, opened 2017-06-19
Avoid polling between sidebar and handle conflicts.

### LINKED ISSUE #71: Polling for changes is annoying
If you take more than a second to select a piece of text and edit or style it, you will loose your selection.
This is pretty annoying and it is due to the polling + refresh strategy.

I already have an idea of how to fix that (like looking at the content and refresh only if it changed) but I also filed a [Bugzilla issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1373621) to see if we can come up with a better strategy for this Sidebar API.

Natim:
Fix #71 #73 

Implements https://docs.google.com/document/d/1aZRxPiTtMgS-WRczxf_DGKr2Dag2J88kj1B1RA87Y0s/edit

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -1,8 +1,7 @@
 const client = new KintoClient('https://kinto.dev.mozaws.net/v1');
 
-const SYNC_BROWSER_SIDEBARS_CONTENT_SECONDS_FREQUENCY = 1.5;
 const LOAD_FROM_KINTO_SECONDS_FREQUENCY = 60;
-const STORE_TO_KINTO_SECONDS_FREQUENCY = 30;
+const STORE_TO_KINTO_SECONDS_FREQUENCY = 10;
 
 const formats = [
   'bold',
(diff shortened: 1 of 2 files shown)


src/sidebar/panel.js
```diff
@@ -1,8 +1,7 @@
 const client = new KintoClient('https://kinto.dev.mozaws.net/v1');
 
-const SYNC_BROWSER_SIDEBARS_CONTENT_SECONDS_FREQUENCY = 1.5;
 const LOAD_FROM_KINTO_SECONDS_FREQUENCY = 60;
-const STORE_TO_KINTO_SECONDS_FREQUENCY = 30;
+const STORE_TO_KINTO_SECONDS_FREQUENCY = 10;
```
Natim: 10 seconds after a change.

src/sidebar/panel.js
```diff
@@ -155,101 +154,123 @@ function loadContent() {
...
+                    content,
+                    result.last_modified
                   );
                 })
                 .catch(err => {
-                  handleLocalContent(data);
                   console.error(err);
+                  Promise.resolve(
+                    browser.storage.local
+                      .set({ lastModified: result.last_modified })
```
Natim: In case we cannot decrypt we define the current local value as the remote value. We need to make sure here that the Bearer token is invalid in case the key changed.

vladikoff:
Had trouble with this:

Profile 1: sign in , write `notes`
Profile 2: sign in, wait for `notes`, but never arrived

vladikoff:
Closing for now, as we are reming the sync code for now

PR #90, opened 2017-06-19
adjust line height on sync note

vladikoff:


vladikoff:
based on feedback from @ryanfeeley

PR #91, opened 2017-06-20
Fix font size menu.

### LINKED ISSUE #28: font sizes don't match
image

See https://github.com/quilljs/quill/issues/1208#issuecomment-303767338

(
Also see https://github.com/quilljs/quill/issues/491
And maybe https://github.com/quilljs/quill/issues/1493
)

Natim:
Fixes #28

### DIFF
--- a/src/sidebar/panel.html
+++ b/src/sidebar/panel.html
@@ -13,11 +13,11 @@
   <div id="toolbar">
     <!-- Add font size dropdown -->
     <select class="ql-size">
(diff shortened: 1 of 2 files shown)


Natim:
I am wondering if we should add 24px and 36px.

Natim:
> I am wondering if we should add 24px and 36px.

@ryanfeeley Any idea?

vladikoff:
Thanks @Natim !

Issue #42, opened 2017-06-14
Add localization support


vladikoff:
See a good example: https://github.com/bwinton/SnoozeTabs

vladikoff:
bug `flod` about L10N

flodolo:
This is one confusing issue subject :-) I'd suggest "Add localization support".

I'm happy to answer questions if you have, I took a quick look at the repository and couldn't spot any string, so you might check at other repositories to see how they manage extracting strings to .properties, etc.

vladikoff:
Sorry, I typed it during the meeting. Updated

flodolo:
Some links in the meantime
https://www.npmjs.com/package/pontoon-to-webext

File structure requested by Pontoon
https://developer.mozilla.org/en-US/docs/Mozilla/Implementing_Pontoon_in_a_Mozilla_website

Pontoon can commit in master or in a branch, like it happens for SnoozeTabs.

Is this a TestPilot experiment?

vladikoff:
> Is this a TestPilot experiment?

Yeap!

Natim:
- [ ] Use `browser.i18n.getMessage('slug')}` everywhere we need to show a translated string.
- [ ] Add `pontoon-to-webext` as a `dev depedency`
- [ ] [List supported locale in packages.json](https://github.com/bwinton/SnoozeTabs/blob/1765637da1e1ed2f3aaf9f801458f1ae179f6d8f/package.json#L63)
- [ ] Create a [./bin/build-locales.js](https://github.com/bwinton/SnoozeTabs/blob/1765637da1e1ed2f3aaf9f801458f1ae179f6d8f/bin/build-locales.js) script and plug it in the `npm run build` step.
- [ ] Create [a locales directory](https://github.com/bwinton/SnoozeTabs/tree/1765637da1e1ed2f3aaf9f801458f1ae179f6d8f/locales) wih translations for supported locales.
- [ ] Then plug Ponton to keep locales translations in sync
- [ ] Create a [`locales/en-US/notes.properties`](https://github.com/bwinton/SnoozeTabs/blob/1765637da1e1ed2f3aaf9f801458f1ae179f6d8f/locales/en-US/snoozetabs.properties) with American translations
[...]

PR #95, opened 2017-06-20
WIP : Add localization support

udaraweerasinghege:
For https://github.com/mozilla/notes/issues/42

- [x] Add pontoon-to-webext as a dev depedency
- [x] Create a locales/en-US/notes.properties with American translations
- [x] Create a ./bin/build-locales.js script and plug it in the npm run build step.
- [ ] Create a locales directory with translations for supported locales.
- [ ]  Use browser.i18n.getMessage('slug')} everywhere we need to show a translated string.
- [ ] Try to avoid hardcoded list of locales!!! OR List supported locale in packages.json
- [ ] Add README to this repo, example: https://github.com/mozilla/activity-stream/blob/master/docs/v1-test-pilot/localization.md
- [ ] Then plug Ponton to keep locales translations in sync
- [ ] Handle langDir ltr, rtl and Configure Quill.js

### DIFF
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -1,0 +1,13 @@
+welcomeTitle=Welcome!
+welcomeText=This is a simple one-page notepad built in to Firefox that helps you get the most out of the web.
+
+emptyPlaceHolder=Take a note…
+
+openingLogin=Opening Login…
+forgetEmail=Forget this Email
+
+syncNotReady=Sorry, Sync is not ready yet, but we will count your click as a vote to hurry it up!
+syncNotes=Sync Your Notes
+syncProgress=Syncing Changes…
+disableSync=Disable Sync
+syncComplete=Synced
--- a/scripts/build-locales.js
+++ b/scripts/build-locales.js
@@ -1,0 +1,18 @@
+#! /usr/bin/env node
+const spawn = require('cross-spawn');
+
+const locales = '*';
+
+const result = spawn.sync('pontoon-to-webext', [], {
+  stdio: 'inherit',
+  env: Object.assign(
+    {},
+    process.env,
+    { SUPPORTED_LOCALES: locales }
+  )
+});
+
+if (result.error) {
+  console.error(result.error); // eslint-disable-line no-console
+  process.exit(1);
+}
--- a/package.json
+++ b/package.json
@@ -11,7 +11,9 @@
     "testpilot-metrics": "2.1.3"
   },
   "devDependencies": {
+    "cross-spawn": "^5.1.0",
     "eslint": "^4.0.0",
+    "pontoon-to-webext": "^1.0.2",
     "prettier": "^1.4.4",
     "rimraf": "^2.6.1",
     "web-ext": "^1.9.1"


locales/en-US/notes.properties
```diff
@@ -0,0 +1,13 @@
+welcomeTitle=Welcome!
+welcomeText=This is a simple one-page notepad built in to Firefox that helps you get the most out of the web.
+
+quillPlaceHolder=Take a note...
```
flodolo: Use … (single unicode character)

locales/en-US/notes.properties
```diff
@@ -0,0 +1,13 @@
...
+
+openingLogin=Opening Login...
```
flodolo: Same here

locales/en-US/notes.properties
```diff
@@ -0,0 +1,13 @@
...
+forgetEmail=Forget this Email
+
+syncNotReady=Sorry, Sync is not ready yet, but we will count your click as a vote to hurry it up!
+syncNotes=Sync Your Notes
+syncProgress=Syncing Changes...
```
flodolo: Same here

locales/en-US/notes.properties
flodolo: Use … (single unicode character)

.gitignore
```diff
@@ -2,4 +2,4 @@ node_modules/
 web-ext-artifacts/
 src/vendor/*
 src/sidebar/vendor/*
-
+dist/*
```
vladikoff: What is dist for? Also I wonder if we need to ignore `_locales` dir...

scripts/build-locales.js
```diff
@@ -0,0 +1,22 @@
+#! /usr/bin/env node
+const spawn = require('cross-spawn');
+
+const locales = '*';
```
Natim: 👍

Natim:
Thank you for starting the work 🥇

vladikoff:
from chatting in IRC we wanted to make sure we land the initial strings, rest of TODO into a different PR

locales/en-US/notes.properties
```diff
@@ -0,0 +1,13 @@
+welcomeTitle=Welcome!
+welcomeText=This is a simple one-page notepad built in to Firefox that helps you get the most out of the web.
+
+emptyPlaceHolder=Take a note…
+
+openingLogin=Opening Login…
+forgetEmail=Forget this Email
+
+syncNotReady=Sorry, Sync is not ready yet, but we will count your click as a vote to hurry it up!
```
flodolo: Side question when trying to localize these strings: is Sync intended as generic synchronization, or Firefox Sync? Same for disableSync later in the file

Natim: It is intended as generic synchronization.

vladikoff: Oh? is there a question for @ryanfeeley ?

vladikoff: I wonder if we should say `Notes Sync`?

flodolo: The string has been exposed with a note saying "Sync" = generic synchronization

flodolo: On a side note, I would really stay away from "Notes Sync", then you need to figure out if it's a brand, if it's localizable, which part given that "Notes" won't be. Basically a can of worms ;-)

ryanfeeley: Well spotted! I think we should say:

syncNotReady=Sorry, syncing notes isn’t ready. We will counted your click as a vote to hurry it up!

I added a non-breaking space between the last two words to avoid an "orphan" text wrap. Keep it in if possible.

PR #96, opened 2017-06-20
feat(translations): call browser i18n for all strings

udaraweerasinghege:
For https://github.com/mozilla/notes/issues/42

TODO: get ltr/rtl working properly

Question:
- Should the name of the add-on be translated? i.e `notes`?

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -26,5 +26,7 @@ You can build the add on with a different locale by using the --pref argument to
 ```WEB_EXT_PREF=general.useragent.locale=locale_to_test npm start```
 
 ## Testing with right to left
-You can test right to left locales by building with a locale that uses RTL text direction. (e.g ar)
+You can run ```npm run start-rtl``` to start the addon with the ar locale
+
+You can also test with a custom RTL locale of your choice with
 ```WEB_EXT_PREF=general.useragent.locale=ar npm start```
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -6,9 +6,7 @@ emptyPlaceHolder=Take a note…
 openingLogin=Opening Login…
 forgetEmail=Forget this Email
 
-# LOCALIZATION NOTE (syncNotReady): Sync is intended as a generic
-# synchronization, not Firefox Sync.
-syncNotReady=Sorry, Sync is not ready yet, but we will count your click as a vote to hurry it up!
+syncNotReady=Sorry, syncing notes isn’t ready. We will count your click as a vote to hurry it up!
 syncNotes=Sync Your Notes
 syncProgress=Syncing Changes…
 # LOCALIZATION NOTE (disableSync): Sync is intended as a generic
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "license": "MPL-2.0",
   "repository": "mozilla/notes",
   "scripts": {
-    "build": "web-ext build -s src && unzip -l web-ext-artifacts/*.zip",
+    "build": "node scripts/build-locales && web-ext build -s src && unzip -l web-ext-artifacts/*.zip",
     "clean": "rimraf web-ext-artifacts",
     "format": "prettier 'src/**/!(vendor)/*.{css,js}' --single-quote --write",
     "lint": "eslint src",
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "postformat": "npm run lint -- --fix",
     "postinstall": "scripts/postinstall.sh",
     "prebuild": "npm run clean",
-    "start": "web-ext run -s src"
+    "start": "web-ext run -s src",
+    "start-rtl": "WEB_EXT_PREF=general.useragent.locale=ar npm start"
   }
 }
--- a/src/sidebar/panel.html
+++ b/src/sidebar/panel.html
@@ -26,6 +26,7 @@
     <button class="ql-strike"></button>
     <button class="ql-list" value="ordered"></button>
     <button class="ql-list" value="bullet"></button>
+    <span id="ql-direction"></span>
   </div>
 
   <!-- Create the Quill.js editor container -->
--- a/src/sidebar/panel.html
+++ b/src/sidebar/panel.html
@@ -33,8 +33,8 @@
 
   <!-- Add a Firefox Sync toolbar -->
   <footer>
-    <div id="sync-note">Sorry, Sync is not ready yet, but we will count your click as a vote to hurry it up!</div>
-    <button id="enable-sync">Sync Your Notes</button>
+    <div id="sync-note"></div>
+    <button id="enable-sync"></button>
   </footer>
 
   <!-- Include the Quill library -->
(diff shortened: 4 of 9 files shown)


vladikoff:
> TODO: get ltr/rtl working properly

Yea, also would be really awesome to figure out how to `npm start` into `rtl` mode right away

locales/en-US/notes.properties
```diff
@@ -6,7 +6,7 @@ emptyPlaceHolder=Take a note…
 openingLogin=Opening Login…
 forgetEmail=Forget this Email
 
-syncNotReady=Sorry, Sync is not ready yet, but we will count your click as a vote to hurry it up!
+syncNotReady=Sorry, syncing notes isn’t ready. We will count your click as a vote to hurry it up!
```
flodolo: You need to rebase this, and at this point drop the localization note. It should be OK to change without changing the string ID, since the meaning is the same, it's just a minor nit.

src/sidebar/panel.js
```diff
@@ -37,6 +42,9 @@ function handleLocalContent(data) {
       quill.setContents(data.notes);
     }
   }
+  const contentLength = quill.getLength();
+  quill.formatText(0, contentLength, 'direction', LANG_DIR);
+  quill.formatText(0, contentLength, 'align', LANG_DIR === 'rtl' ? 'right' : 'left');
```
udaraweerasinghege: this is a workaround as `setContents` with the attribute `direction` doesnt seem to work as expected. https://github.com/quilljs/quill/issues/1535

src/sidebar/panel.js
```diff
@@ -6,12 +6,17 @@ const formats = [
...
   'header',
   'indent',
-  'list'
+  'list',
+  'direction',
+  'align'
 ];
+const UI_LANG = browser.i18n.getUILanguage();
+const RTL_LANGS = ['ar', 'fa', 'he'];
+const LANG_DIR = RTL_LANGS.indexOf(UI_LANG) === -1 ? 'ltr' : 'rtl';
```
pdehaan: Nit: you can use the ES6-ish [`Array#includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) instead of checking against -1 index.

package.json
```diff
@@ -22,7 +22,7 @@
   "license": "MPL-2.0",
   "repository": "mozilla/notes",
   "scripts": {
```
vladikoff: @udaraweerasinghege @pdehaan do we have a way to do something like:  `RTL=true npm start` or `LOCALE=he npm start` via the web-ext tool?

udaraweerasinghege: Since as you found we can do `WEB_EXT_PREF=general.useragent.locale=fr-FR`, should I just add that to the readme?

udaraweerasinghege:
Updated checklist
- [x] Use browser.i18n.getMessage('slug')} everywhere we need to show a translated string.
- [x] Add pontoon-to-webext as a dev depedency
- [x] Try to avoid hardcoded list of locales!!! OR List supported locale in package.json
- [x] Create a ./bin/build-locales.js script and plug it in the npm run build step.
- [x] Create a locales directory wih translations for supported locales.
- [x] Then plug Ponton to keep locales ranslations in sync
- [x] Create a locales/en-US/notes.properties with American translations
- [x] Handle langDir ltr, rtl and Configure Quill.js
- [ ]  Add README to this repo, example: https://github.com/mozilla/activity-stream/blob/master/docs/v1-test-pilot/localization.md

package.json
vladikoff: @udaraweerasinghege Let's add a script command as well. Maybe `npm run start-rtl` that would use the `ar`locale for now

vladikoff:
@udaraweerasinghege Looks good, testing now....

Can we build locales after `npm install` is done in `postinstall`? this way if I `npm start` it would not fail.

udaraweerasinghege:
cc: @vladikoff added building locales into postinstall and added a start-rtl script

src/sidebar/panel.js
```diff
@@ -26,14 +33,14 @@ function handleLocalContent(data) {
   if (!data.hasOwnProperty('notes')) {
     quill.setContents({
       ops: [
-        { attributes: { size: 'large', bold: true }, insert: 'Welcome!' },
-        { insert: '\n\n' },
+        { attributes: { size: 'large', bold: true }, insert: browser.i18n.getMessage('welcomeTitle') },
+        { insert: '\n\n', attributes: { direction: LANG_DIR, align: TEXT_ALIGN_DIR }},
```
Natim: This is not going to work, it might for the initial text but then if you type text, you need somehow to tell quill.js about the text direction too.

src/sidebar/panel.html
```diff
@@ -26,6 +26,7 @@
     <button class="ql-strike"></button>
     <button class="ql-list" value="ordered"></button>
     <button class="ql-list" value="bullet"></button>
+    <span id="ql-direction"></span>
```
vladikoff: @udaraweerasinghege can we make this a `button` like the rest above?

vladikoff: nvm

src/sidebar/panel.js
udaraweerasinghege: Since we're adding the direction on the newline at the end it did hold the formatting for the next new line as well as you type. There was a weird behaviour though where if you select all the text and type over it, it goes back to LTR. However, if you select all, delete text, and type it'll be good in RTL. We've added an icon so that in the last case they can just press the icon to change it back to RTL.

in the gif the first time i select all -> delete -> type
and the second time i select all -> type

ql

docs/localization.md
```diff
@@ -26,5 +26,7 @@ You can build the add on with a different locale by using the --pref argument to
 ```WEB_EXT_PREF=general.useragent.locale=locale_to_test npm start```
 
 ## Testing with right to left
-You can test right to left locales by building with a locale that uses RTL text direction. (e.g ar)
+You can run ```npm run start-rtl``` to start the addon with the ar locale
```
vladikoff: nice!

src/sidebar/panel.js
```diff
@@ -6,16 +6,29 @@ const formats = [
   'strike',
   'header',
   'indent',
-  'list'
+  'list',
+  'direction',
+  'align'
```
vladikoff: @udaraweerasinghege should we remove `align` ?

src/sidebar/panel.js
```diff
@@ -6,16 +6,29 @@ const formats = [
...
 ];
+const UI_LANG = browser.i18n.getUILanguage();
+const RTL_LANGS = ['ar', 'fa', 'he'];
+const LANG_DIR = RTL_LANGS.includes(UI_LANG) ? 'rtl' : 'ltr';
+const TEXT_ALIGN_DIR = LANG_DIR === 'rtl' ? 'right' : 'left';
```
vladikoff: same as https://github.com/mozilla/notes/pull/96/files#r123545949?

src/sidebar/panel.js
udaraweerasinghege: Align must be allowed for direction to work, as shown below.

Without align: 
image


With align
image

vladikoff: Got it, thanks!

vladikoff:
Thanks @udaraweerasinghege

PR #97, opened 2017-06-20
Adds draft of metrics documentation.

chuckharmston:
Note: the link to the Test Pilot standards will 404 while in review; for now, it can be read at https://github.com/mozilla/testpilot/blob/2538-ga-docs/docs/experiments/ga.md.

CC @johngruen, @clouserw

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,113 @@
+# Notes Metrics
+The metrics collection and analysis plan for Notes, a forthcoming Test Pilot experiment.
+
+
+## Analysis
+Data collected by Notes will be used to answer the following high-level questions:
+
+- Do people take notes in Firefox?
+	- In what intervals?
+	- In what circumstances?
+- What are the notes like?
+	- How long are they?
+	- Do they span multiple lines?
+	- Do they use formatting options? Which?
+- Do people sync?
+	- How does syncing affect engagement?
+
+Note: sync will not be part of the first release of this experiment; all sync-related metrics will be reported when that feature is offered.
+
+## Collection
+Data will be collected with Google Analytics and follow [Test Pilot standards](https://github.com/mozilla/testpilot/blob/master/docs/experiments/ga.md) for reporting.
+
+### Custom Metrics
+- `cm1` - the size of the notepad, in characters, with markup removed.
+- `cm2` - the size of the notepad, measured by the number of lines.
+- `cm3` - the size of the change, measured in the number of characters different between the content of the notepad when they began editing and the contents when the `change` event was fired, with markup removed.
+
+### Custom Dimensions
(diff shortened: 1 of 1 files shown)


docs/metrics.md
```diff
@@ -0,0 +1,113 @@
...
+- `cd1`
+- `cd2`
+- `cd3`
+- `cd4`
+- `cd5`
+- `cd6`
+- `cd7`
+
+#### `changed`
+An event fired when the user completes a change of the content of the notepad. It prospectively begins when a user focuses on the notepad's editable area, and ends when the user either 1) closes the sidebar, or 2) does not make any changes in 120 seconds. Includes:
```
chuckharmston: Moved from https://github.com/mozilla-services/hoverpad/pull/55#discussion_r123113260:

120s seems long, but I don't know what would be better. maybe jgruen has opinions

docs/metrics.md
```diff
@@ -0,0 +1,113 @@
...
+- Do people sync?
+	- How does syncing affect engagement?
+
+Note: sync will not be part of the first release of this experiment; all sync-related metrics will be reported when that feature is offered.
+
+## Collection
+Data will be collected with Google Analytics and follow [Test Pilot standards](https://github.com/mozilla/testpilot/blob/master/docs/experiments/ga.md) for reporting.
+
+### Custom Metrics
+- `cm1` - the size of the notepad, in characters, with markup removed.
```
Natim: Can we use more explicit names than cm1, cm2, etc?

What we had here: https://gist.github.com/chuckharmston/adb9df1aaed2d769b2a3ff1a41f766c8 with the pad and user object seems much more easier to understand and is self documented.

Natim: Here is the current getPad info function: https://github.com/mozilla/notes/pull/98/files

docs/metrics.md
Natim: I guess that value could be the same as syncing remotely to Kinto. 20 seconds after a change?

docs/metrics.md
chuckharmston: Unfortunately, we're stuck with those names; that's just how GA's custom dimensions and metrics work. Some links:

- https://github.com/mozilla/testpilot/blob/master/docs/experiments/ga.md#custom-dimensionmetric-fields
- https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#customs

docs/metrics.md
chuckharmston: Happy for this to be changed to whatever makes sense.

docs/metrics.md
Natim: Ok makes sense then :)

wresuolc:
Thanks Chuck!

PR #117, opened 2017-06-29
Clean src/sidebar/vendor/ directory

pdehaan:
Take it or leave it, but I just realized I've been building the add-on locally with a bunch of bloated old vendor files which we aren't using anymore (which was doubling my extension size).

This should clean and rebuild the ./src/sidebar/vendor/* directory.


Natim:
Nice catch thanks.

PR #128, opened 2017-07-10
update strings

### LINKED ISSUE #30: string review on the onboarding string
from meeting: @johngruen will send it to review

johngruen:
Fixes #30

### DIFF
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -1,12 +1,12 @@
-welcomeTitle=Welcome!
-welcomeText=This is a simple one-page notepad built in to Firefox that helps you get the most out of the Web.
+welcomeTitle2=Hello!
+welcomeText2=Welcome to this one-page notepad built in to Firefox. Browse the web. Jot down notes. It’s that simple.
 
 emptyPlaceHolder=Take a note…
 
 openingLogin=Opening Login…
 forgetEmail=Forget this Email
 
-syncNotReady=Sorry, syncing notes isn’t ready. We will count your click as a vote to hurry it up!
+syncNotReady2=Sorry, syncing notes quite isn’t ready. We will count your click as a vote to hurry it up!
 syncNotes=Sync Your Notes
 syncProgress=Syncing Changes…
 # LOCALIZATION NOTE (disableSync): Sync is intended as a generic
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -55,12 +55,12 @@ function handleLocalContent(data) {
   if (!data.hasOwnProperty('notes')) {
     quill.setContents({
       ops: [
-        { attributes: { size: 'large', bold: true }, insert: browser.i18n.getMessage('welcomeTitle') },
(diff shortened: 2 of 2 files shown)


locales/en-US/notes.properties
```diff
@@ -1,12 +1,12 @@
-welcomeTitle=Welcome!
-welcomeText=This is a simple one-page notepad built in to Firefox that helps you get the most out of the Web.
+welcomeTitle2=Hello!
+welcomeText2=Welcome to this one-page notepad built in to Firefox. Browse the web. Jot down notes. It's that simple.
```
flodolo: It's -> It’s

johngruen: You and @ryanfeeley are eagle-eyed about this! Thanks.

Natim: What is the difference between `It's` and `It’s` `'` and `´` ?

flodolo: For strings you should always use proper apostrophes/curly quotes instead of straight single/double quotes. 

That's the rule within Firefox copy, to the point that you would fail tests if you try to land a strings with ' or "", instead of ’ and “”

locales/en-US/notes.properties
```diff
@@ -1,12 +1,12 @@
...
+welcomeTitle2=Hello!
+welcomeText2=Welcome to this one-page notepad built in to Firefox. Browse the web. Jot down notes. It’s that simple.
 
 emptyPlaceHolder=Take a note…
 
 openingLogin=Opening Login…
 forgetEmail=Forget this Email
 
-syncNotReady=Sorry, syncing notes isn’t ready. We will count your click as a vote to hurry it up!
+syncNotReady2=Sorry, syncing notes quite isn’t ready. We will count your click as a vote to hurry it up!
```
vladikoff: @johngruen `notes quite isn’t ready` vs `notes isn’t quite ready`?

ryanfeeley: My 2¢: "notes isn’t quite ready" is correct.

locales/en-US/notes.properties
ryanfeeley: # It’s nice to come back from PTO and see that, with respect to typographic punctuation, it’s handled correctly @flodolo.

src/sidebar/panel.js
```diff
@@ -55,12 +55,12 @@ function handleLocalContent(data) {
...
     quill.setContents({
       ops: [
-        { attributes: { size: 'large', bold: true }, insert: browser.i18n.getMessage('welcomeTitle') },
+        { attributes: { size: 'large', bold: true }, insert: browser.i18n.getMessage('welcomeTitle2') },
         { insert: '\n\n', attributes: { direction: LANG_DIR, align: TEXT_ALIGN_DIR }},
         {
           attributes: { size: 'large' },
           insert:
-            browser.i18n.getMessage('welcomeText')
+            browser.i18n.getMessage('welcomeText2')
```
Natim: When we change ids like that, shouldn't we fallback on the previous id value? Especially for languages that haven't the new string translation yet?

PR #131, opened 2017-07-10
Handle breaklines on shift-enter and copy/paste from a text area.

### LINKED ISSUE #127: Pasting inserts extra spaces
screen shot 2017-07-07 at 4 20 09 pm

becomes
```
notSupportedHeader = Your browser is not supported.

notSupportedHeader = Unfortunately this browser does not support the web technology that powers Firefox Send. You'll need to try another browser. We recommend Firefox!

// Note Firefox name should not be localized in button, but subhead should

downloadFirefoxButtonSub = Free Download

// note the sentence lower on this page is the same as uploadPageExplainer. Also the link on this page should be removed as Send wont work on this browser anyway
```

when copy pasted

Natim:
Fixes #127

### DIFF
--- a/src/sidebar/styles.css
+++ b/src/sidebar/styles.css
@@ -45,6 +45,10 @@ div#editor {
   margin-left: -5px;
 }
 
+.ql-editor p {
+	margin-bottom: 10px;
+}
+
 /**
 Footer Styles
  */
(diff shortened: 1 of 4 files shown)


src/sidebar/styles.css
```diff
@@ -45,6 +45,10 @@ div#editor {
   margin-left: -5px;
 }
 
+.ql-editor p {
+	margin-bottom: 10px;
```
vladikoff: a lot of indent here

Natim:
Superseeded by #148

PR #133, opened 2017-07-11
Sync button styling (hover and active)

### LINKED ISSUE #129: Style Sync button with hover and active states
Button should have visible states.

cedricium:
Fixes #129. Added styling for the `:hover` and `:active` states. Colors used
are based off the Photon UI visual design guide (1).

The button wasn't actually flush with the bottom of the Notes panel so I
moved it downwards with a negative margin value. I then compensated by adding
the same amount of padding to the bottom of the button to keep the contents
of the button horizontally centered.

1. https://mozilla.invisionapp.com/share/PYC5LJJXG#/screens/229940646


Natim:
out

@ryanfeeley Does it looks good to you?

Natim:
out

Adding the click too.

vladikoff:
@cedricium 

We need to make this look like the proposed design here: 

image

cedricium:
Got it! Will make the necessary changes :+1:

cedricium:
Here's what the new changes look like:

syncbtn-styling

Let me know what you think!

vladikoff:
@cedricium Hey! just 2 quick things:

image

1. Can we add a border on top of the yellow warning?, see 

2. Can we make the padding on top and bottom of the message smaller, so it fits the design more:

cedricium:
Still a WIP, will let you know when it's ready to be looked at.

ryanfeeley:
@cedricium I've removed the scaling in Invision so this mockup is 1:1 and should be pixel perfect to copy. Want red lines? I think Sketch can generate them for me.
https://mozilla.invisionapp.com/share/6VBUYHMRB#/235284916_Desktop_Sidebar

cedricium:
@ryanfeeley @vladikoff I made the changes based off the Invision designs, let me know what you think.

vladikoff:
@cedricium looks almost ready to go, sadly forgot to mention, there is a bit of a radius on the button: 

image

It's a little hard to see, grey vs white radius of the button

ryanfeeley:
@vladikoff That's the worse screenshot I ever seen. 
image

vladikoff:
@ryanfeeley thanks Ryan!

cedricium:
hey @vladikoff, if you look at the `#enablesync:hover` and `#enablesync:active` lines in `style.css` (I need to learn how to reference code in comments), you'll see I used an inset box-shadow to achieve the highlighting when hovering and clicking on the sync button.

I've tried playing around with the code and after a bit of researching, I'm not sure how I can get the corners to be rounded. Adding a  `border-radius` to the button doesn't work and only gives the button itself rounded corners but not the `:hover` and `:active` highlighting.

vladikoff:
@cedricium I guess we will need to add margins and maybe make the button nested in the source

cedricium:
Made quite a few changes but it seems to work well and looks nearly identical to the Invision mockups. Would definitely want as many different machines/browsers tested just to make sure.

pdehaan:
> I need to learn how to reference code in comments

@cedricium Whether you want to cite a specific line in a random file or in a PR, the process is pretty much the same.
You can go to a random file (like [/scripts/postinstall.sh](https://github.com/mozilla/notes/blob/master/scripts/postinstall.sh#L3)) and click a line number and it should change the URL to something like this: https://github.com/mozilla/notes/blob/master/scripts/postinstall.sh#L3 (also note that the clicked line should highlight, which is pretty neat).

That is relatively useful, but not entirely. Note that the link still points to **master** branch. Now, if somebody else modifies that file and the line numbers change, your comment/link would point to line 3 in the modified file which may not be exactly what you were referencing earlier. To get around this, you can press the "Y" key and it should give you a modified URL like https://github.com/mozilla/notes/blob/4210c0454f7fb5a0fb9a9973d1c9677bee994ccb/scripts/postinstall.sh#L3 which is for the latest/current Git SHA. Super cool! Now if somebody else modifies the file, it will always point to that specific line.

Oh, also note that you can get somewhat creative and press the Shift key and create a range of selected lines: https://github.com/mozilla/notes/blob/4210c0454f7fb5a0fb9a9973d1c9677bee994ccb/.eslintrc.yml#L9-L14. Although I've never found a way to select+highlight multiple non-consecutive lines.

And it should be fairly similar for a PR: https://github.com/mozilla/notes/pull/134/files#diff-42ce51f6d4dfb013156ee9d7209291f6R118, although not sure if the "Y" trick works there, but maybe the weird "#diff-{SHA}" URL hash solves that "changing file contents" thing.

cedricium:
Wow, thanks so much @pdehaan! The multi-line highlighting doesn't appear to work on commits, but here's a shot at it anyways:

https://github.com/mozilla/notes/pull/133/commits/49ec75f815290df18a020e2c0a1bef2a43e3ce2e#diff-42ce51f6d4dfb013156ee9d7209291f6L73-L76

vladikoff:
Great work @cedricium , thank you so much!

PR #134, opened 2017-07-11
Use a no-drop pointer when dragging into the ql-editor.

### LINKED ISSUE #132: [Enhancement] Mouse pointer should reflect that action is not allowed when trying to drop content into sidebar
[Notes]:
- Since the drag and drop feature was disabled, the cursor should also reflect that this action is not allowed when the user tries to drop content into Firefox Notes.

[Affected versions]:
- Firefox 53.0 and up
- Firefox Notes dev v1.5

[Affected Platforms]:
- All Windows
- All Linux
- All Mac

[Prerequisites]:
- Have a Firefox profile with the latest Firefox Notes add-on version (1.5 -dev, built on 10/07/2017) installed.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Navigate to a webpage and drag any content over the Firefox Notes sidebar.
3. Observe the mouse pointer icon.

[Expected result]:
- The mouse pointer changes when Notes sidebar is hovered to reflect that the drag and drop action is not allowed.

[Actual result]:
- The mouse pointer changes into a "drop" icon instead, even if the action is not performed when released.

[Additional Notes]:
mousepoiner_issue

Natim:
Fixes #132 

Apparently it doesn't work properly. I filed https://github.com/quilljs/quill/issues/1582 about it.

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -122,7 +122,19 @@ chrome.runtime.onMessage.addListener(eventData => {
 
 // disable drop of links and images into notes
 const qlEditor = document.querySelectorAll('.ql-editor');
-qlEditor[0].addEventListener('drop', (e) => {
+
+document.addEventListener('dragover', () => {
+  qlEditor[0].classList.add('forbid-cursor');
+  return true;
+});
+
+document.addEventListener('dragleave', () => {
+  qlEditor[0].classList.remove('forbid-cursor');
+  return true;
+});
+
+document.addEventListener('drop', (e) => {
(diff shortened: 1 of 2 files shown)


Natim:
After discussion on the Quill.js front, we discovered that it also happens to be a bug with textarea themselves.

A bug has been open here: https://bugzilla.mozilla.org/show_bug.cgi?id=1381759

Natim:
I am going to land that so that it will operate as soon as the bug is fixed on Firefox side.

PR #141, opened 2017-07-13
Bring back FxA login using the crypto relier library

vladikoff:
- [x] Split the Kinto related function to a specific file to keep background.js smaller.

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -3,6 +3,10 @@
  */
 const TRACKING_ID = 'UA-35433268-79';
 
+// XXX: Read this from Kinto fxa-params
+const FXA_CLIENT_ID = 'c6d74070a481bc10';
+const FXA_OAUTH_SERVER = 'https://oauth-scoped-keys.dev.lcip.org/v1';
+
 const timeouts = {};
 
 const analytics = new TestPilotGA({
@@ -36,36 +40,37 @@ function sendMetrics(event, context = {}) {
   timeouts[event] = setTimeout(later, 20000);
 }
 
+function authenticate() {
+  const fxaKeysUtil = new FxaCrypto.relier.OAuthUtils();
+
+  fxaKeysUtil.launchFxaScopedKeyFlow({
+    client_id: FXA_CLIENT_ID,
+    oauth_uri: FXA_OAUTH_SERVER,
+    pkce: true,
+    redirect_uri: browser.identity.getRedirectURL(),
+    scopes: ['profile', 'https://identity.mozilla.org/apps/notes'],
+  }).then((loginDetails) => {
+    console.log('access token + keys', loginDetails);
+    chrome.runtime.sendMessage({
+      action: 'authenticated',
+      bearer: loginDetails.access_token,
+      keys: loginDetails.keys
+    });
+  }, (err) => {
+    console.log('login failed', err);
+    chrome.runtime.sendMessage({
+      action: 'authenticated',
+      err: err
+    });
+    throw err;
+  });
+}
+
 browser.runtime.onMessage.addListener(function(eventData) {
   switch (eventData.action) {
     case 'authenticate':
       sendMetrics('webext-button-authenticate', eventData.context);
-
-      const fxaKeysUtil = new FxaCrypto.relier.OAuthUtils();
-
-      fxaKeysUtil.launchFxaScopedKeyFlow({
-        client_id: 'c6d74070a481bc10',
-        //oauth_uri: 'http://127.0.0.1:9010/v1',
-        oauth_uri: 'https://oauth-scoped-keys.dev.lcip.org/v1',
-        pkce: true,
-        redirect_uri: browser.identity.getRedirectURL(),
-        scopes: ['profile', 'https://identity.mozilla.org/apps/notes'],
-      }).then((loginDetails) => {
-        console.log('access token + keys', loginDetails);
-        chrome.runtime.sendMessage({
-          action: 'authenticated',
-          bearer: loginDetails.access_token,
-          keys: loginDetails.keys
-        });
-
-      }, (err) => {
-        console.log('login failed', err);
-        chrome.runtime.sendMessage({
-          action: 'authenticated',
-          err: err
-        });
-        throw err;
-      });
+      authenticate();
       break;
     case 'metrics-changed':
       sendMetrics('changed', eventData.context);
--- a/src/background.js
+++ b/src/background.js
@@ -48,7 +48,7 @@ browser.runtime.onMessage.addListener(function(eventData) {
 
       fxaKeysUtil.launchFxaScopedKeyFlow({
         client_id: 'c6d74070a481bc10',
-        oauth_uri: 'http://127.0.0.1:9010/v1',
+        oauth_uri: 'https://oauth-scoped-keys.dev.lcip.org/v1',
         pkce: true,
         redirect_uri: browser.identity.getRedirectURL(),
         scopes: ['profile', 'https://identity.mozilla.org/apps/notes'],
--- a/src/background.js
+++ b/src/background.js
@@ -39,17 +39,14 @@ function sendMetrics(event, context = {}) {
 browser.runtime.onMessage.addListener(function(eventData) {
   switch (eventData.action) {
     case 'authenticate':
-      sendEvent({
-        object: 'webext-button-authenticate',
-        method: 'click'
-      });
+      sendMetrics('webext-button-authenticate', eventData.context);
 
       const fxaKeysUtil = new FxaCrypto.relier.OAuthUtils();
 
       fxaKeysUtil.launchFxaScopedKeyFlow({
         client_id: 'c6d74070a481bc10',
-        oauth_uri: 'http://127.0.0.1:9010/v1',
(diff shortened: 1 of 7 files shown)


src/sidebar/panel.js
```diff
@@ -170,3 +169,17 @@ qlEditor[0].addEventListener('drop', (e) => {
   e.preventDefault();
   return false;
 });
+
+chrome.runtime.onMessage.addListener(eventData => {
+  switch (eventData.action) {
+    case 'authenticated':
+      if (eventData.err) {
+        enableSync.textContent = 'Login Failed...';
```
flodolo: Don't you want this and the next message localized? Also, … -> ...

vladikoff: yea thanks. this is just some work in progress

vladikoff:
## This was rebased

vladikoff:
Rebased this just now again

src/background.js
```diff
@@ -3,8 +3,19 @@
  */
 const TRACKING_ID = 'UA-35433268-79';
 
+const KINTO_SERVER = 'https://kinto.dev.mozaws.net/v1';
+// XXX: Read this from Kinto fxa-params
+const FXA_CLIENT_ID = 'c6d74070a481bc10';
+const FXA_OAUTH_SERVER = 'https://oauth-scoped-keys.dev.lcip.org/v1';
```
pdehaan: Should these be hardcoded vs pulled/build from some ENV vars?
Seems odd that we'd have a "production-esque" add-on distributed via Test Pilot talking to dev instances of services.

vladikoff: The pr won't be merged until this is using prod urls :) it's gonna be using a prod kinto url

src/manifest.json
```diff
@@ -22,11 +22,16 @@
...
   "permissions": [
     "storage",
+    "identity",
     "https://ssl.google-analytics.com/collect"
   ],
   "background": {
     "scripts": [
+      "vendor/jose.js",
+      "vendor/kinto-http.js",
+      "vendor/fxa-crypto-relier/fxa-crypto-relier.js",
```
pdehaan: See https://github.com/mozilla/notes/pull/141/files#r139488935 comment, re: do we need that extra `/fxa-crytpo-relier/` directory with 1 file?

src/sidebar/panel.js
```diff
@@ -201,14 +233,68 @@ function getThemeFromStorage() {
     }
   });
 }
+
+function getLastSyncedTime() {
+  const getting = browser.storage.local.get(['last_modified', 'credentials']);
+  getting.then(data => {
+    if (data.hasOwnProperty('credentials')) {
+      const time = new Date(data.last_modified).toLocaleTimeString();
+      enableSync.textContent = 'Synced at ' + time;
```
pdehaan: Should this be localized?

src/sidebar/panel.js
```diff
@@ -201,14 +233,68 @@ function getThemeFromStorage() {
...
+    }
+  });
+}
+
 document.addEventListener('DOMContentLoaded', getThemeFromStorage);
+document.addEventListener('DOMContentLoaded', getLastSyncedTime);
```
pdehaan: Should we have two listeners for the event, or just a single listener which calls the two functions?

For example:

```js
document.addEventListener('DOMContentLoaded', () => {
  getThemeFromStorage(),
  getLastSyncedTime()
});
```

src/sidebar/panel.js
```diff
@@ -201,14 +233,68 @@ function getThemeFromStorage() {
...
+  let time;
+  let content;
   switch (eventData.action) {
+    case 'sync-authenticated':
+      chrome.runtime.sendMessage({
+          action: 'kinto-load'
+        });
+      break;
+    case 'kinto-loaded':
+      console.log("kinto-loaded content", eventData);
```
pdehaan: I think this needs to use single quotes... ¯\\\_(ツ)\_/¯

src/sync.js
```diff
@@ -0,0 +1,165 @@
...
+  const later = function() {
+    syncDebounce = null;
+
+    browser.storage.local.get(['credentials', 'notes', 'contentWasSynced', 'last_modified'])
+      .then(data => {
+        if (!data.contentWasSynced && data.hasOwnProperty('credentials')) {
+          return encrypt(data.credentials.key, data.notes)
+            .then(encrypted => {
+              const headers = { Authorization: `Bearer ${data.credentials.access_token}`};
+              if (data.hasOwnProperty('last_modified') && typeof data.last_modified !== "object") {
```
pdehaan: Needs single quotes...

src/sidebar/panel.js
```diff
@@ -201,14 +233,68 @@ function getThemeFromStorage() {
...
+          content = newContent.ops.concat(local.ops);
+        } else {
+          content = remote;
+        }
+        ignoreNextTextChange = true;
+      } else {
+        browser.storage.local.remove('initialContent');
+        content = eventData.data;
+      }
+    browser.storage.local.set({ notes: content, last_modified: eventData.last_modified });
```
pdehaan: Indenting looks funny here...

src/sidebar/panel.js
```diff
@@ -201,14 +233,68 @@ function getThemeFromStorage() {
...
+    case 'text-editing':
+      enableSync.textContent = 'Editing';
+      break;
+    case 'text-synced':
+      time = new Date(eventData.last_modified).toLocaleTimeString();
+      enableSync.textContent = 'Synced at ' + time;
+      break;
+    case 'text-saved':
+      time = new Date().toLocaleTimeString();
+      enableSync.textContent = 'Saved at ' + time;
```
pdehaan: Needs localization?

src/sidebar/panel.js
```diff
@@ -201,14 +233,68 @@ function getThemeFromStorage() {
...
+      break;
     case 'theme-changed':
       getThemeFromStorage();
```
pdehaan: Nit: is this a case without a `break`? That could cause some sketchy bugs in the future. ⚠️

src/sync.js
```diff
@@ -0,0 +1,165 @@
...
+                    .collection('notes')
+                    .deleteRecord('singleNote', {
+                      headers: { Authorization: `Bearer ${data.credentials.access_token}` }
+                    });
+                }
+              });
+          }
+        })
+        .catch(error => {
+          if (/HTTP 404/.test(error.message)) {
```
pdehaan: Not sure if we want some more friendly wrappers for these, such as `is404()`, `is401() {}`, `is412() {}`, etc.

or maybe something even more generic:

```js
function checkStatus(msg, code) {
  return new RegExp(`HTTP ${code}`).test(msg);
}

if (checkStatus(error.message, 404)) {
  ...
}

// function is404(msg) {
//  return checkStatus(msg, 404);
// }
```

src/sync.js
```diff
@@ -0,0 +1,165 @@
...
+    syncDebounce = null;
+
+    browser.storage.local.get(['credentials', 'notes', 'contentWasSynced', 'last_modified'])
+      .then(data => {
+        if (!data.contentWasSynced && data.hasOwnProperty('credentials')) {
+          return encrypt(data.credentials.key, data.notes)
+            .then(encrypted => {
+              const headers = { Authorization: `Bearer ${data.credentials.access_token}`};
+              if (data.hasOwnProperty('last_modified') && typeof data.last_modified !== "object") {
+                headers['If-Match'] = '"' + data.last_modified + '"';
```
pdehaan: Not sure if you maybe want template strings here:

```js
headers['If-Match'] = `"${data.last_modified}"`;
```

src/sync.js
```diff
@@ -0,0 +1,165 @@
...
+          } else {
+            console.error(error);
+          }
+        });
+    });
+}
+
+function saveToKinto(client) {
+  // XXX: Debounce the call and set the status to Editing
+  browser.runtime.sendMessage('notes@mozilla.com', {
```
pdehaan: I don't pretend I understand the message sending stuff, but I'll _assume_ that we're OK with it sometimes being `chrome.runtime.sendMessage()` vs `browser.runtime.sendMessage()` and sometimes `browser.runtime.sendMessage('notes@mozilla.com', {...})` vs other times where we don't specify the `'notes@mozilla.com'` as the first arg.

src/background.js
pdehaan: @vladikoff: OK, cool... but how will that be configured? ENV var? Hard coded?
That means that if we do load testing on this in the future, any load testery would be hitting a prod server endpoint (unless we have flags or we remember to manually edit the endpoint and rebuild add-on locally).

Natim: Should we put that in the config?

src/sidebar/panel.js
Natim: I like to have two so that it is easiser to remove one eventually.

src/sync.js
Natim: Yes we specify the ID of the addons when we want to send the event to all open sidebars rather than when we just want to send it to the current one.

The only benefit of using chrome instead of browser is if we want to support to run the add-on in Chrome, which is not the case since we added Firefox Account.

Natim:
Superseeded by #241

PR #149, opened 2017-07-18
Reduce startup states.

### LINKED ISSUE #137: Reduce startup states
This slowed-down clip of switching to Notes shows about six different states. Is there a way that we can reduce this to one?
[slowmo.mp4.zip](https://github.com/mozilla/notes/files/1143208/slowmo.mp4.zip)

Natim:
Wait for a full loading before showing the editor. Fixes #137

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -132,6 +132,7 @@ function loadContent() {
 
 loadContent()
   .then(() => {
+    document.getElementById('loading').style.display = 'none';
     browser.runtime.sendMessage({
       action: 'metrics-open',
       context: getPadStats()
--- a/src/sidebar/styles.css
+++ b/src/sidebar/styles.css
@@ -15,6 +15,16 @@ body {
   padding: 0;
 }
 
+div#loading {
+	height: 100%;
+	width: 100%;
+	position: absolute;
+	top: 0;
+	left: 0;
+	background: white;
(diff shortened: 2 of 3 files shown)


src/sidebar/styles.css
```diff
@@ -15,6 +15,16 @@ body {
   padding: 0;
 }
 
+div#loading {
+	height: 200%;
```
Natim: can probably use 100% here.

src/sidebar/panel.js
```diff
@@ -121,13 +121,19 @@ function handleLocalContent(data) {
 }
 
 function loadContent() {
-  browser.storage.local.get(['notes'], data => {
-    // If we have a bearer, we try to save the content.
-    handleLocalContent(data);
+  return new Promise((resolve) => {
+    browser.storage.local.get(['notes'], data => {
```
vladikoff: Can this fail? Can we log to GA that it failed?

src/sidebar/styles.css
```diff
@@ -15,6 +15,16 @@ body {
   padding: 0;
 }
 
+div#loading {
+	height: 100%;
```
vladikoff: indent seems wrong here

src/sidebar/panel.js
Natim: No it cannot fail.

PR #174, opened 2017-07-27
Add l10n.toml

flodolo:
This allows to run compare-locales in the Notes repo, check for errors in locales

```
compare-locales l10n.toml . `ls locales/`
```

### DIFF
--- a/l10n.toml
+++ b/l10n.toml
@@ -1,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+basepath = "."
+
+[env]
+    l = "{l10n_base}/locales/{locale}/"
+
+[[paths]]
+    reference = "locales/en-US/**"
+    l10n = "{l}**"


flodolo:
CC @pike

pdehaan:
A few other notes since I'm annoying...

For anybody else with no l10n experience and curious how to install `compate-locales`:

```sh
$ sudo easy_install -U compare-locales
```
Ref: https://developer.mozilla.org/en-US/docs/Mozilla/Projects/compare-locales

Also, not sure if ```compare-locales l10n.toml . `ls locales/` ``` super user friendly, or if we want to try wrapping it in an npm script so we can just run `$ npm run compare_locales`. 🤷‍♀️

Pike:
There's a `locales` entry in l10n.toml which lists the default list of localizations.

We haven't figured out a way to make that the single source of truth, right now, pontoon just commits new locales, assuming that someone picks them up.

http://moz-l10n-config.readthedocs.io/en/latest/fileformat.html has docs, https://hg.mozilla.org/comm-central/file/tip/mail/locales/l10n.toml an example.

flodolo:
It might be useful to update that list of locales at some point, but given how early it is for this add-on, I want to wait a bit more before doing it.

PR #190, opened 2017-08-04
Add Keyboard Shortcut

### LINKED ISSUE #176: Consider a keyboard shortcut to open the notes sidebar?
I can open history with ctrl-h and bookmarks with ctrl-b. It would be great if I could open notes with alt-n or something.

vladikoff:
Fixes #176

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -36,7 +36,7 @@
   "commands": {
     "_execute_sidebar_action": {
       "suggested_key": {
-        "default": "Ctrl+Shift+H"
+        "default": "Alt+Shift+N"
       }
     }
   }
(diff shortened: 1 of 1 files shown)


Natim:
I also tried `Ctrl+Shift+N` without luck. @flodolo shall we plan on localize the shortcut?

flodolo:
> I also tried Ctrl+Shift+N without luck. @flodolo shall we plan to localize the shortcut?

I'm not sure. This is a shortcut, not an acccesskey, so normally locales follow English.

If it opens the Edit menu, it means you're triggering accesskeys instead of a shortcut (like CMD+T to open a new tab on Mac).

Natim:
Let's go with `Ctrl+Shift+L` which seems to be the first available letter...

Natim:
@vladikoff suggested `Alt+Shift+W` which seems to work as well.

PR #202, opened 2017-08-14
Convert bash script to node script

pdehaan:
Fixes #200

PR #206, opened 2017-08-14
Fixes issue #205(Improvement in README document)

### LINKED ISSUE #205: Improvement needed In README document
New users can get quite confused as steps mentioned to set up the development environment are not clear
and also some commands are not working as expected(in windows).Therefore improvement is needed.

Divya063:
Fixes #205


package.json
```diff
@@ -32,7 +32,7 @@
     "lint:webext": "web-ext lint -s src --ignore-files='sidebar/vendor/*.js' --self-hosted",
     "package": "npm run build && mv web-ext-artifacts/*.zip addon.xpi",
     "postformat": "npm run lint:js -- --fix",
-    "postinstall": "scripts/postinstall.sh",
+    "postinstall": "bash ./scripts/postinstall.sh",
```
vladikoff: We cannot assume that all developers use `bash`

README.md
```diff
@@ -18,6 +23,13 @@ Open a notepad in the Firefox toolbar to store notes about your browsing.
 | `npm run build` | Builds the application as a Web Extension.
 | `npm start`     | Launches Firefox with the Web Extension.
 
+NOTE: if you are running on Windows, and experience a build error use `$ npm build`
+
+Step 3:Load about:debugging#addons in the Firefox address bar.
+
+Step 4:Select a Firefox .xpi file or Manifest file under the src folder and click open afterwards
```
vladikoff: You don't have to do this, `npm start` automatically installs the extension using the `web-ext` tool

README.md
```diff
@@ -9,7 +9,12 @@ Open a notepad in the Firefox toolbar to store notes about your browsing.
 
 
 ## How to use it
-
+Step 1: Make sure you fork and clone [notes](https://github.com/mozilla/notes).
+```
+$ git clone https://github.com/[yourusername]/notes.git
+```
+Step 2: Install its dependencies
```
vladikoff: Do we need instructions to `npm install` here?

vladikoff:
Thanks for the Pull Request, left some notes!

README.md
```diff
@@ -18,6 +23,13 @@ Open a notepad in the Firefox toolbar to store notes about your browsing.
 | `npm run build` | Builds the application as a Web Extension.
 | `npm start`     | Launches Firefox with the Web Extension.
 
+NOTE: if you are running on Windows, and experience a build error use `$ npm build`
```
pdehaan: I _think_ this needs to be `npm run build` here.

Divya063: `npm run buid` is giving an error 
screenshot 201
`npm build` is working fine on windows 
screenshot 203

README.md
Divya063: Thank you for the review, yes it is not needed. For some reason I had to load it manually...removing those lines.Thanks again

package.json
Natim: Refs #202

Divya063: sh or bash works either way when I'm using the git bash shell.However, this works only for windows users.Thanks for the update @Natim

xfix: Seems to be JavaScript script now

README.md
Natim: Can you try again `npm run build` with this new update? Here (on linux) `npm build` doesn't do anything.

Also if it fails with `npm run build` on windows it should fail with `npm build` as well because the error is not related to npm but to windows.

Issue #166, opened 2017-07-24
Add CHANGELOG.md to track differences between releases?


pdehaan:
We may just be able to run diffs against releases/tags, but that can be tedious: https://github.com/mozilla/notes/tags

Natim:
Released in 1.7.0 https://github.com/mozilla/notes/issues/188

PR #209, opened 2017-08-17
Add Changelog md

vladikoff:
Fixes https://github.com/mozilla/notes/issues/166


CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+<a name="1.7.0"></a>
```
Natim: I don't think we need this html tag here.

Natim: Maybe we want to start with `# Changelog`

PR #212, opened 2017-08-18
Add release steps.

Natim:


vladikoff:
@Natim how does this work for point releases, see https://github.com/mozilla/notes/issues/214 ?

vladikoff:
from meeting: We also need to add a step for QA: "Send release to QA a day before publishing..."

Natim:
> Send release to QA a day before publishing...

How does this works? Are we creating a branch that they can try out?

Natim:
> how does this work for point releases?

It works exactly the same way I guess.

vladikoff:
Thanks @Natim !!

PR #224, opened 2017-09-01
Add @ryanfeeley SVG icon.

### LINKED ISSUE #217: Notes extension icon is not pixel snapped
I was pretty sure that this was provided at a pixel snapped size (so that right angles are not anti-aliased), but in Nightly it's been scaled/moved a little. How to fix?
screen shot 2017-08-24 at 3 23 26 pm

Natim:
Fixes #217


ryanfeeley:
@Natim Cool! Can I get a screenshot?

Natim:
capture

Natim:
Maybe it is better if you can zoom it yourself to see if it fixes it because I can just tell you that it looks like the correct new icon

ryanfeeley:
Darn. It does not. You are using Ubuntu? Those lines should be perfect white lines. Also I set the fill of the object to be "context-fill" so it should inherit the theme colour (and match the other icons in being more grey).

Any ideas?

ryanfeeley:
Darn. It does not. You are using Ubuntu? Those lines should be perfect white lines. Also I set the fill of the object to be "context-fill" so it should inherit the theme colour (and match the other icons in being more grey).

Any ideas?

Natim:
> You are using Ubuntu?

Yes

ryanfeeley:
@Natim Can you tell me what the exact size is required? The SVG is sized to 32×32px but will pixel snap down to 16×16px.

Natim:
Apparently that's the issue, the if you open the svg with Firefox and scale it down three time you will see the exact same rendering. I guess we should make sure we can display the icon in 16x16, 22x22, 32x32 and 48x48

16x16: scale-five-times

22x22: scale-three-times

32x32: normal-size

Natim:
capture

Here is how it looks with the new icon. (I made sure the space and their size where divisible by 4

Natim:
New update capture

ryanfeeley:
@Natim It's hard to tell as your screenshot is not pixel perfect. @vladikoff Any ideas?

Natim:
> It's hard to tell as your screenshot is not pixel perfect.

@ryanfeeley Don't hesitate to try the branch on your own computer ;)

Issue #208, opened 2017-08-17
Add a saving status indicator


Natim:
Tell people that their notes were saved.

Siddhinita:
@Natim I propose adding a snackbar to alert the user. What do you think?

vladikoff:
@Siddhinita we have a design almost ready

Siddhinita:
@Natim Is it already implemented?If not, can I work on it?

ryanfeeley:
Will we have a time stamp? If yes, will it look different than, for example:

* Thursday 3:45 PM

(my locale)

Natim:
@ryanfeeley we can have a timestamp if you'd like to

Natim:
@Siddhinita When you think about a snackbar are you thinking of something like that?

![snackbar](https://storage.googleapis.com/material-design/publish/material_v_11/assets/0B6Okdz75tqQsZV9TS3lTVHFzRUE/components_snackbar_usage_fabdo_005.webm)

Siddhinita:
@Natim..Yes.

ryanfeeley:
How is this? Text colour is #808080
![saving-indicator](https://user-images.githubusercontent.com/68213/29462458-6dac57a6-83fd-11e7-8e17-dfec28af3b28.gif)

Natim:
I kind of like it, how will it work with the syncing then?

ryanfeeley:
@Natim [It would work the same.](https://mozilla.invisionapp.com/share/6VBUYHMRB#/235284916_Desktop_Sidebar) The save status indicator becomes the sync status indicator. The sync icon moves to the left and is replaced with a ⋯ button that lets user disable syncing.

ryanfeeley:
If we have access to a time stamp,  use locale's equivalent of: 
    Saved Thursday 3:45 PM

johngruen:
Todo: me
https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#experiments

vladikoff:
@cedricium Hey! do you have time to add the menu functionality for this?

![](http://v14d.com/i/59c1209ae82ca.jpg)

Following this design but the menu will have multiple items. One of them is "Feedback".

cedricium:
@vladikoff yea, I'll get started right away! Besides "Feedback", what would be the other menu items?

Natim:
From top to bottom:

- Feedback
- Disable Sync

vladikoff:
[...]

PR #240, opened 2017-09-19
Context Menu for Footer

cedricium:
This adds the context menu for #208. Currently, the menu looks like so:
screen shot 2017-09-19 at 11 06 05 am

Work that still needs to be completed: 
- [x] dynamically load menu items (see next todo) - **completed: aa65927**
- [x] use translations instead of hard-coded strings for menu items - **completed: aa65927**
- [x] `eventListener`'s for menu items:
  - [x] 'Feedback' - **completed: 4f7f72d**
  ~- [ ] 'Disable Sync'~
- [x] styling for dark theme - **completed: 8775c0b**
- [x] styling that more closely matches @ryanfeeley's design
- [x] clean up code styling (tabs -> spaces; empty lines) - **completed: 23e0989**

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "url": "https://github.com/mozilla/notes/issues"
   },
   "dependencies": {
-    "material-design-lite": "^1.3.0",
+    "material-design-lite": "1.3.0",
     "quill": "1.2.6",
     "testpilot-ga": "^0.2.1"
   },
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "url": "https://github.com/mozilla/notes/issues"
   },
   "dependencies": {
+    "material-design-lite": "^1.3.0",
     "quill": "1.2.6",
     "testpilot-ga": "^0.2.1"
   },
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -8,7 +8,8 @@ const files = [
   copySync('node_modules/quill/dist/quill.snow.css', 'src/sidebar/vendor/quill.snow.css'),
   copySync('node_modules/quill/LICENSE', 'src/sidebar/vendor/quill.LICENSE'),
   copySync('node_modules/material-design-lite/material.min.js', 'src/sidebar/vendor/material.js'),
-  copySync('node_modules/material-design-lite/material.min.css', 'src/sidebar/vendor/material.css')
+  copySync('node_modules/material-design-lite/material.min.css', 'src/sidebar/vendor/material.css'),
+  copySync('node_modules/material-design-lite/LICENSE', 'src/sidebar/vendor/material.LICENSE')
 ];
 
 Promise.all(files).catch(err => {
--- a/scripts/postinstall.js
+++ b/scripts/postinstall.js
@@ -6,7 +6,9 @@ const files = [
   copySync('node_modules/testpilot-ga/dist/index.js', 'src/vendor/testpilot-ga.js'),
   copySync('node_modules/quill/dist/quill.min.js', 'src/sidebar/vendor/quill.js'),
   copySync('node_modules/quill/dist/quill.snow.css', 'src/sidebar/vendor/quill.snow.css'),
-  copySync('node_modules/quill/LICENSE', 'src/sidebar/vendor/quill.LICENSE')
+  copySync('node_modules/quill/LICENSE', 'src/sidebar/vendor/quill.LICENSE'),
+  copySync('node_modules/material-design-lite/material.min.js', 'src/sidebar/vendor/material.js'),
+  copySync('node_modules/material-design-lite/material.min.css', 'src/sidebar/vendor/material.css')
 ];
 
 Promise.all(files).catch(err => {
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -22,6 +22,7 @@
   },
   "permissions": [
     "storage",
+    "tabs",
     "https://ssl.google-analytics.com/collect"
   ],
   "background": {
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -22,7 +22,6 @@
   },
   "permissions": [
     "storage",
-    "tabs",
     "https://ssl.google-analytics.com/collect"
   ],
   "background": {
--- a/src/sidebar/panel.html
+++ b/src/sidebar/panel.html
@@ -4,10 +4,8 @@
   <meta charset="utf-8" />
   <title>Firefox Notes</title>
   <link href="vendor/quill.snow.css" rel="stylesheet">
+  <link rel="stylesheet" href="vendor/material.css">
   <link href="styles.css" rel="stylesheet">
-
-	<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
-	<link rel="stylesheet" href="vendor/material.css">
 </head>
 
 <body>
@@ -45,17 +43,13 @@
 
     <div id="footer-buttons">
       <button id="enable-sync"></button>
-			<div class="wrapper">
-				<button id="demo-menu-top-right"
-        class="mdl-js-button">
-					<i class="material-icons">more_vert</i>
-				</button>
-				<ul class="mdl-menu mdl-menu--top-right mdl-js-menu"
-    data-mdl-for="demo-menu-top-right">
-					<li id="give-feedback" class="mdl-menu__item"></li>
-					<li id="disable-sync" class="mdl-menu__item"></li>
-				</ul>
-			</div>
+      <div class="wrapper">
+        <button id="context-menu-button" class="mdl-js-button"></button>
+        <ul class="mdl-menu mdl-menu--top-right mdl-js-menu context-menu" data-mdl-for="context-menu-button">
+          <li id="give-feedback" class="mdl-menu__item context-menu-item"></li>
+          <li id="disable-sync" class="mdl-menu__item context-menu-item"></li>
+        </ul>
+      </div>
     </div>
   </footer>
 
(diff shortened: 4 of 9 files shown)


vladikoff:
Awwww yesss!

scripts/postinstall.js
```diff
@@ -6,7 +6,9 @@ const files = [
   copySync('node_modules/testpilot-ga/dist/index.js', 'src/vendor/testpilot-ga.js'),
   copySync('node_modules/quill/dist/quill.min.js', 'src/sidebar/vendor/quill.js'),
   copySync('node_modules/quill/dist/quill.snow.css', 'src/sidebar/vendor/quill.snow.css'),
-  copySync('node_modules/quill/LICENSE', 'src/sidebar/vendor/quill.LICENSE')
+  copySync('node_modules/quill/LICENSE', 'src/sidebar/vendor/quill.LICENSE'),
+  copySync('node_modules/material-design-lite/material.min.js', 'src/sidebar/vendor/material.js'),
+  copySync('node_modules/material-design-lite/material.min.css', 'src/sidebar/vendor/material.css')
```
pdehaan: Hhmm, do we need to copy the **material-design-lite**'s LICENSE file here as well?

src/sidebar/panel.html
```diff
@@ -5,6 +5,9 @@
   <title>Firefox Notes</title>
   <link href="vendor/quill.snow.css" rel="stylesheet">
   <link href="styles.css" rel="stylesheet">
+
+	<link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
pdehaan: Indenting on these 2 new lines looks a bit off...

Also, possible to bundle the font family into the add-on? It seems unfortunate to have to proxy everything from the 3rd party fonts.googleapis.com site.

/cc @vladikoff

src/sidebar/panel.html
```diff
@@ -42,13 +45,24 @@
 
     <div id="footer-buttons">
       <button id="enable-sync"></button>
-      <a target="_blank" id="give-feedback"></a>
+			<div class="wrapper">
+				<button id="demo-menu-top-right"
+        class="mdl-js-button">
+					<i class="material-icons">more_vert</i>
```
pdehaan: Sorry, I'm not super familiar w/ Material, but what is "more_vert" here? I imagine it'd possibly cause some issues for screen readers, unless the CSS is tweaking a bunch of ARIA settings or something.

src/sidebar/styles.css
```diff
@@ -109,6 +110,7 @@ footer {
   display: flex;
   flex: 0 0 40px;
 }
+*/
```
pdehaan: Are we not using this selector anymore? Should we just ✂️ it?

src/sidebar/styles.css
```diff
@@ -177,3 +179,7 @@ footer {
 .forbid-cursor {
   cursor: no-drop;
 }
+
+.mdl-js-button {
+  height: 100%;
+}
```
pdehaan: Again, sorry,  not super familiar w/ Material Design, but this feels pretty global and specific and could cause some future 🔥.

Should we add a more generic class name here as well so we can style _just_ that one element instead of all buttons globally (even though we only use one at the moment)?

```jsx
<button id="demo-menu-top-right" class="mdl-js-button more-button">
  <i class="material-icons">more_vert</i>
</button>
```

src/sidebar/panel.html
```diff
@@ -42,13 +45,24 @@
 
     <div id="footer-buttons">
       <button id="enable-sync"></button>
-      <a target="_blank" id="give-feedback"></a>
+			<div class="wrapper">
+				<button id="demo-menu-top-right"
```
pdehaan: Indenting here looks a bit funny too.

Also, should the button's ID be a bit more... informative (or removed if we're not using it)?

pdehaan:
As always, awesome work @cedricium! Thanks for all the help!

src/sidebar/panel.js
```diff
@@ -156,14 +156,23 @@ quill.on('text-change', () => {
...
 const closeButton = document.getElementById('close-button');
 enableSync.textContent = browser.i18n.getMessage('syncNotes');
 syncNoteBody.textContent = browser.i18n.getMessage('syncNotReady2');
-giveFeedback.setAttribute('title', browser.i18n.getMessage('giveFeedback'));
-giveFeedback.setAttribute('href', SURVEY_PATH);
+
+const giveFeedback = document.getElementById('give-feedback');
+giveFeedback.textContent = browser.i18n.getMessage('feedback');
+giveFeedback.addEventListener('click', () => {
+  browser.tabs.create({
```
vladikoff: Can we avoid adding "tabs" permission?

cedricium: For as simple as it seems, I couldn't get the `giveFeedback` element (`<li>`) to open a **new tab** in the **same window**. I was able to do the following, which none seem correct for opening the survey in terms of UX:

- same window, current tab (user's current tab gets overridden)
- new window w/ one tab (happened when `<a target="_blank">` was nested inside `<li>` element) (breaks user's workflow w/ having a new window open)

I may have missed something though so please let me know if you know of a better way.

cedricium:
@pdehaan thanks for the quick review! I started off with my IDE setup to use tabs so that's the reason for the weird spacing. I'll definitely add material's (MDL) LICENSE. As for selectors, I agree, will change over to targeting specific elements with id's as opposed to the global element.

And for `more_vert`, that's a specific icon (the 3 stacked dots). Since we're only using that, I could download that resource and drop the `<link href="https://fonts.googleapis.com/icon?family=Material+Icons">` altogether.

cedricium:
After the latest commit, this is what the menu looks like:
context-menu

src/sidebar/panel.js
vladikoff: >  I couldn't get the giveFeedback element (<li>) to open a new tab in the same window.

Don't worry about changing this. the way the current feedback window opens is fine. We cannot introduce a new permission just for this.

package.json
```diff
@@ -8,6 +8,7 @@
     "url": "https://github.com/mozilla/notes/issues"
   },
   "dependencies": {
+    "material-design-lite": "^1.3.0",
```
vladikoff: Let's lock this to just `1.3.0`

src/manifest.json
```diff
@@ -22,6 +22,7 @@
   },
   "permissions": [
     "storage",
+    "tabs",
```
vladikoff: let's remove this

src/sidebar/panel.html
```diff
@@ -4,6 +4,7 @@
   <meta charset="utf-8" />
   <title>Firefox Notes</title>
   <link href="vendor/quill.snow.css" rel="stylesheet">
+  <link rel="stylesheet" href="vendor/material.css">
```
vladikoff: to keep the same code style as below put "href" first then "rel".  Also can we put it BEFORE the quill CSS so hopefully nothing in `material` overrides `quill`

src/sidebar/panel.html
```diff
@@ -42,13 +43,20 @@
 
     <div id="footer-buttons">
       <button id="enable-sync"></button>
-      <a target="_blank" id="give-feedback"></a>
+      <div class="wrapper">
+        <button id="context-menu-button" class="mdl-js-button"></button>
+        <ul class="mdl-menu mdl-menu--top-right mdl-js-menu context-menu" data-mdl-for="context-menu-button">
+          <li id="give-feedback" class="mdl-menu__item context-menu-item"></li>
+          <li id="disable-sync" class="mdl-menu__item context-menu-item"></li>
```
vladikoff: Let's remove this button and just have 1 "Feedback" button. the "sync" button will be added as part of a different PR.

src/sidebar/styles.css
vladikoff: 👍

src/sidebar/styles.css
```diff
@@ -177,3 +179,25 @@ footer {
...
+  font-family: inherit;
+  padding: 0;
+}
+
+.context-menu-item {
+  font-size: 12px;
+  padding: 0 0 0 15px;
+  height: 18px;
+  line-height: 30px;
+}
```
vladikoff: need a new line at the of the file

cedricium:
@vladikoff should I change the color of the Feedback button for the dark theme or leave it as is?
screen shot 2017-09-20 at 8 50 58 am

ryanfeeley:
@cedricium It looks like [Pocket switches menus](https://sriramramani.wordpress.com/2012/12/06/runtime-theme-change/) on Android dark theme. I suppose we should too? Do you have the colors you need?

cedricium:
@ryanfeeley should I use the dark theme colors that Notes is using or use the colors outlined in that article?

vladikoff:
> use the dark theme colors that Notes is using 

is correct 👍  :)

cedricium:
Having one issue with completing the dark theme for the menu. When the menu is opened, the menu item flashes it's original whitish color before the dark background is applied:
gif

@vladikoff any ideas on how to fix this? Been toying around all day and I can't find a solution.

vladikoff:
@cedricium is it the background that flashes, or is it the effect of opening the menu? maybe you have to modify more CSS, the amation CSS. I will check it out in a bit.

cedricium:
@vladikoff it appears to me like it's the `.mdl-menu__container` element. Try inspecting [this page](https://getmdl.io/components/index.html#menus-section) and playing around with the DOM, you'll see what I mean. I think you're right that it involves an the animation.

cedricium:
@vladikoff I think I got it! How's this?
gif

vladikoff:
@cedricium looks good! 👍

cedricium:
I'm going to go ahead and say this PR is no longer a WIP and good to go after a final review.

vladikoff:
Thanks @cedricium !! 👍

PR #241, opened 2017-09-19
Introduce kinto

glasserc:
By popular request, an attempt to introduce kinto.js offline-first syncing to try to simplify conflict handling. This just uses the bog standard IndexedDB "storage adapter", which works fine in web extensions, and happens to have a cleaner implementation than browser.storage.local anyhow.

This PR is a bit of a mess since I just plopped a bunch of code into place, but some additional information is in the commit messages. I tried to at least mark the dangerous places with FIXMEs.


vladikoff:
awwww yessss

src/background.js
```diff
@@ -12,7 +12,7 @@ const timeouts = {};
 
 // Kinto sync and encryption
 
-const client = new KintoClient(KINTO_SERVER);
+const client = new Kinto({remote: KINTO_SERVER, bucket: "default"});
```
pdehaan: Should the bucket name be configurable?

src/sync.js
```diff
@@ -30,136 +30,201 @@ function decrypt(key, encrypted) {
...
+    super("key used to encrypt the record appears to be older than our key");
+  }
+}
+
+class JWETransformer {
+  constructor(key) {
+    this.key = key;
+  }
+
+  async encode(record) {
```
pdehaan: I think ESLint is choking here and failing to parse the entire file... Should this be `async function encode(record) {`?

src/sync.js
```diff
@@ -30,136 +30,201 @@ function decrypt(key, encrypted) {
...
+      kid: this.key.kid,
+    };
+    if (record.hasOwnProperty("last_modified")) {
+      encryptedResult.last_modified = record.last_modified;
+    }
+
+    return encryptedResult;
+  }
+
+  async decode(record) {
```
pdehaan: Same here.. should this be `async decode(record) {}` or `async function decode(record) {}`?

src/background.js
pdehaan: Also, I think we need to convert from double quotes to single quotes and declare Kinto as a global to appease the ESLint gods:

```sh
/home/travis/build/mozilla/notes/src/background.js
  15:20  error    'Kinto' is not defined        no-undef
  15:57  error    Strings must use singlequote  quotes
```

src/sync.js
```diff
@@ -30,136 +30,201 @@ function decrypt(key, encrypted) {
   });
 }
 
+// An "id schema" used to validate Kinto IDs and generate new ones.
+const notesIdSchema = {
+  // FIXME: Maybe this should generate IDs?
+  generate() {
+    throw new Error("cannot generate IDs");
```
pdehaan: Once the `encode`/`decode` issues are resolved below, I believe ESLint will start complaining about use of double quotes instead of single quotes here. Sorry.

src/sync.js
```diff
@@ -30,136 +30,201 @@ function decrypt(key, encrypted) {
...
+  }
+
+  async encode(record) {
+    // FIXME: should we try to obfuscate the record ID?
+    const ciphertext = await encrypt(this.key, record);
+    // Copy over the _status field, so that we handle concurrency
+    // headers (If-Match, If-None-Match) correctly.
+    // DON'T copy over "deleted" status, because then we'd leak
+    // plaintext deletes.
+    const status = record._status && (record._status == "deleted" ? "updated" : record._status);
```
pdehaan: Nit: should use `===` instead.

src/sync.js
```diff
@@ -30,136 +30,201 @@ function decrypt(key, encrypted) {
...
+          strategy: "server_wins",
         });
+    })
+    .then(syncResult => {
+      // FIXME: conflicts would happen here.
+      // Do we need to do anything with errors, published, updated, etc.?
+      return syncResult;
+    })
+    .catch(error => {
+      if (error.response && error.response.status == 401) {
```
pdehaan: Nit: use `===` here.

src/background.js
glasserc: I don't think the bucket should be configurable. The default bucket is "API sugar" to allow each user to get their own bucket while still referring to them in a uniform way. It's the only one which behaves like this.

src/sync.js
glasserc: Ah, is it maybe the case that you guys stick to ES2015 on this project? Or which standard do you support? Per http://kangax.github.io/compat-table/es2016plus/#test-async_functions_async_methods,_classes, this syntax is supported in FF 52 ESR and greater, which is why it's used in the storage.sync implementation, which is why I copied and pasted it here :blush:

vladikoff:
@glasserc @Natim Getting `result` is not defined on an existing account: 



Will try a new account now

vladikoff:
Seems to be working for a new account!!!!

vladikoff:
@Natim Seems to working well, i think the biggest thing we need is handling merge conflicts.

There is some minor stuff such as the last sync time not updating: 

I synced at :44 on the hour, but still said :41:



But that's small fish

vladikoff:
@Natim seems like after I close 2 profiles that I tested sync on and opened a 3rd profile I didn't get my notes back (got invalid date): 

image

image

Natim:
First alpha version of the branch: [addon-2.0.0a1.zip](https://github.com/mozilla/notes/files/1371772/addon-2.0.0a1.zip)

Natim:
New version with QuillJS 1.3.3 [addon-2.0.0a2.zip](https://github.com/mozilla/notes/files/1371823/addon-2.0.0a2.zip)

src/sync.js
```diff
@@ -91,6 +91,9 @@ class JWETransformer {
       throw new Error('No ciphertext: nothing to decrypt?');
     }
 
+    // FIXME: this is hack to work around a bug with FxA keys, which
+    // seem to have timestamps embedded in them that are affected by
+    // TZ offsets. Take out the substr calls once vladikoff says we can.
```
Natim: 👍

src/background.js
```diff
@@ -6,7 +6,7 @@ const TRACKING_ID = 'UA-35433268-79';
 const KINTO_SERVER = 'https://kinto.dev.mozaws.net/v1';
 // XXX: Read this from Kinto fxa-params
 const FXA_CLIENT_ID = 'c6d74070a481bc10';
-const FXA_OAUTH_SERVER = 'https://oauth-scoped-keys.dev.lcip.org/v1';
+const FXA_OAUTH_SERVER = 'https://oauth-scoped-keys-oct10.dev.lcip.org/v1';
```
vladikoff: @Natim @glasserc please update the kinto dev server to use this now. If this server is not working for you while I'm on PTO please rollback this commit :)

vladikoff: This has the fixes to the `kid` and other stuff

Natim:
Since this PR is against sync-v4, I am going to land it and then we can keep improving sync-v4 branch until it is ready.

Issue #233, opened 2017-09-18
Add support for autolink (typing and pasting)


Natim:
Refs https://github.com/quilljs/quill/issues/109#issuecomment-219265086

cedricium:
As far as typing goes, I used the code [outlined in this answer](https://github.com/quilljs/quill/issues/109#issuecomment-278181150) and it works pretty well. Based on the regex used in that code, only text beginning with `http://` or `https://` will be recognized, whereas Google Docs matches text beginning with just `www.` as well.

Another note with the example code given in the above comment is that the typed text should be followed by whitespace (either a space or newline will do), otherwise any proceeding text may become a part of the link itself. One way of preventing this would be to implement the <kbd>Cmd/Ctrl</kbd><kbd>K</kbd> shortcut to allow toggling on and off linking for highlighted text.

vladikoff:
> only text beginning with http:// or https:// will be recognized

That sounds good for starters 👍 

> Another note with the example code given in the above comment is that the typed text should be followed by whitespace (either a space or newline will do)

That's also fine. I think G Docs might do this too.

> Do you suggest removing that editing tooltip entirely, or replacing it with our own?  

If we can avoid the tooltip entirely that would be great. Due to the sidebar nature of the project we are trying to stay away from these tooltips :)

cedricium:
Created a pull request with the basic support for links. Still need to add link mouse click events to the analytics.

Natim:
Fixed with 75c751452146e488b39e7b8c86fb38df4bdc3ecb

PR #244, opened 2017-09-25
Add support for links

cedricium:
This is the initial implementation for supporting links (#233). With these changes,
links will automatically be created from anything that begins with `http://`
or `https://`. That includes any typed or pasted text.

The changes in the CSS make it so that the Quill link editing UI tooltip is
no longer shown. In order to open a link, the user must mouse over the link
and right-click for the context menu or <kbd>Ctrl/Cmd</kbd>+<kbd>Left Click</kbd> to open in a new tab.

**What Needs to Be Added:**
- [x] link mouse click events added to analytics - **completed: 877af94**
- [x] cursor mouseover behavior and appearance - **completed: 877af94**


vladikoff:
@cedricium great start!!

Two things I noticed:

## 1. links to do not open

when clicked:

link


## 2. typing out urls, it's not easy to get out of the "URL" styling:

link2

Natim:
For the second bug, the goal would be to remove the link as soon as it is not a valid link anymore.

cedricium:
1. I wasn't sure what the expected behavior should be. Because this is text editor, links are not inherently clickable since editing a notepad is in the 'editing' state, not the 'readable' state. But I can change the behavior and make the links clickable for Notes purpose.
2. Once you add whitespace after a url, the link is created. Hitting `[Backspace]` begins editing the link immediately, whereas Google Docs for example removes the link after the first `[Backspace]` to allow users to make edits. I will go ahead and try to replicate this behavior as well.

vladikoff:
> But I can change the behavior and make the links clickable for Notes purpose.

👍 

> Hitting [Backspace] begins editing the link immediately

Yea with the default UI that we hide there was a "remove" button, but because we are getting rid of it we need a way to get out of link editing. If it is not possible to do then we will need to find another way. (Maybe by bringing back the UI and making it simpler).

cedricium:
@vladikoff let me know if this latest commit fixes the issues, I think it should. Whenever you delete the whitespace character after the link then begin typing, the `link` formatting should be removed. Links are clickable as well.

I tried implementing the `link-clicked` analytics as well, or so I thought. Let me know if I did that correctly.

vladikoff:
@cedricium works really well! Doing so final testing before merging :)

vladikoff:
@cedricium Hey, so the only change we want to make is:

If the link starts with https:// http:// it should edit the link when you edit the text.

Example:
link2

so in the gif above, If i edit the link it should go to `Firefox` instead of `Firefox for iOS`

cedricium:
@vladikoff I'll see what I can do! 👍

Natim:
@cedricium I added some debounce as well as links modification handling. Can you both validate with @vladikoff that it works as expected?

Natim:
Apparently my commit breaks pasting links.

cedricium:
@Natim I removed your debounce work as it seemed to be messing with the functionality of links. If possible, I'd like to have this latest commit sanity tested to verify the issues #252, #254, and #255 were actually resolved. If that's good, then I'd like to add back in debounce and continue working on the rest of the links issues.

cc @vladikoff

Natim:
I am going to land this and we can iterate from it.

vladikoff:
awww yessss , amazing work @cedricium // @Natim 

🎉

Issue #277, opened 2017-10-03
Add option to underline the text at the top panel


RMCoder198:
Add option to underline the text at the top panel .
![mo](https://user-images.githubusercontent.com/26675957/31137317-83a1309a-a888-11e7-9d3b-9f6446f124f0.png)

I would like to work on this.

johngruen:
@RMCoder198 we're going to sit tight on this right now b/c we're doing some refactoring of the toolbar UI this week. i'll keep you looped in to changes as warranted.

vladikoff:
No plans to add the underline button, but shortcut works

Ref: https://mozilla.invisionapp.com/share/6VBUYHMRB#/screens/281041484_Firefox_Notes

PR #278, opened 2017-10-04
Fixes issue #277:Add option to underline the text at the top panel

RMCoder198:
Add option to underline the text at the top panel.

After changes 
underline


vladikoff:
Thanks for the Pull Request! @ryanfeeley is our UX lead, he will review if we want to add this button.

Natim:
We are delaying the merge of this PR because the toolbar min size cannot handle one more icon just yet.

We will add it as soon as we rework the way the toolbar should be responsive.

Issue #253, opened 2017-09-27
Hyperlinks format options are canceled when choosing the numbered (or bulleted) list options


SoftVision-CarmenFat:
[Affected versions]:
- Firefox 55.0.3 and up
- Firefox Notes v1.8.0rc2

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Firefox Notes version (1.8.0rc2).

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Write a valid hyperlink address into the "Notes" sidebar (eg. http://wikipedia.org)
3. Select the hyperlink and click on one of the following formatting buttons: Font Size, Bold, Italic or Strikethrough.
4. Click on Numbered List (or Bulleted List) button and observe the hyperlink.

[Expected result]:
- The hyperlink is displayed as a numbered (or bulleted) list presenting the same formatting option from step 3.

[Actual result]:
- The hyperlink is displayed as a numbered (or bulleted) list, but the previous formatting option is canceled.

[Notes]:
- This issue is not reproducible when having more than one Hyperlinks selected.
- Attached a screen recording of the issue:

![list options](https://user-images.githubusercontent.com/26000308/30915176-9eeb67f6-a39e-11e7-896b-04f801f96308.gif)

Natim:
Fixed in 75c751452146e488b39e7b8c86fb38df4bdc3ecb

SoftVision-CarmenFat:
The issue is still reproducible on the latest Firefox Notes version (1.8.0rc3).

Natim:
@cedricium I think this is due to the fact that we recreate the node. Do you know how we can update it instead?

Natim:
Similar reason for https://github.com/mozilla/notes/issues/269

cedricium:
@Natim I'll need to investigate this, but I suspect you're right about the node being recreated.

SoftVision-CarmenFat:
The issue is no longer reproducible with the latest Firefox Notes v.1.8.0dev (built today 10/24/2017).

Issue #273, opened 2017-10-02
"Space" key wrongly creates linked underline if pressed just before the hyperlink


SoftVision-CarmenFat:
[Affected versions]:
- Firefox 56 and up
- Firefox Notes v1.8.0rc3

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Firefox Notes version (1.8.0rc3).

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Write a valid hyperlink (e.g. http://wikipedia.org).
3. Click on the space right before the hyperlink.
4. Press "Space" key at least 3 times and see what happens next.

[Expected result]:
- The hyperlink is moved 3 spaces to the right.

[Actual result]:
- The hyperlink is moved 3 spaces to the right, but the first two spaces contain underline linked to the website.

[Notes]:
- The issue is not reproducible if plain text was previously written before the hyperlink.
- Attached a screen recording of the issue:

![underline](https://user-images.githubusercontent.com/26000308/31077880-dd636cc0-a788-11e7-8df1-c1756f873370.gif)

vladikoff:
cc @cedricium @Natim

SoftVision-CarmenFat:
- Tested this issue again and I can confirm it is not reproducible following the original steps to reproduce.
- However, a similar behavior is still reproducible when pressing "**Tab**" key twice, instead of "Space" key.

Tested with Firefox Notes RC5.

![again underline](https://user-images.githubusercontent.com/26000308/31282450-438bd200-aabb-11e7-8e86-e33d5685ba17.gif)

cedricium:
@SoftVision-CarmenFat can you verify this when you get a chance, should be in the latest `master` branch.

cedricium:
@Natim this can be closed as well.

SoftVision-CarmenFat:
The issue is no longer reproducible with the latest Firefox Notes v.1.8.0dev (built today 10/23/2017), using both "Space" and "Tab" keys.

Issue #271, opened 2017-10-02
In a formatted text, an inserted or written link doesn't keep active the preexistent formatting


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 56 and up
- Firefox Notes v1.8.0rc3

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Firefox Notes version (1.8.0rc3).

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Have some styled text (Font size, Bold/Italic/Strikethrough style etc.).
3. Type or insert a link in the formatted text.
4. Observe the behavior.

[Expected result]:
- The link is correctly displayed, respecting the preexistent formatting text style.

[Actual result]:
- The link is not correctly displayed and it doesn't respect the preexistent formatting text style.

[Notes]:
- Here is a screen recording of the issue:
![notes - writing issue](https://user-images.githubusercontent.com/31244235/31077219-64b0630c-a786-11e7-8ff3-d031dbafd486.gif)

Softvision-CristinaBadescu:
I was able to reproduce this issue in two different scenarios, using Firefox Notes 1.8.0dev version, on latest Nightly 58.0a1 build (2017-10-19), on Windows 10x64, Mac 10.12 and Ubuntu 14.04.

**First scenario:**

[Steps to reproduce]
1. Write something in "Notes" sidebar, without formatting.
2. Change the font size and select any formatting options.
3. Type a link and observe its behavior.

[Actual results]
- The typed link is not keeping the active text formatting.

Here is a screen recording with this behavior:
![notes - scenario 1](https://user-images.githubusercontent.com/31244235/31823781-daa44bea-b5b5-11e7-97ed-a8e5aa3ba2cf.gif)


**Second scenario:**

[Steps to reproduce]
1. Write something in "Notes" sidebar, without formatting.
2. Change the font size, select any formatting options and paste a link.
3. Start writing something and observe the behavior.

[Actual results]
- The pasted link is not keeping the active text formatting when starting writing something.

Here is a screen recording with this behavior:
![notes - scenario 2](https://user-images.githubusercontent.com/31244235/31823788-e40d3c3c-b5b5-11e7-8011-6c0baabeb585.gif)

Natim:
Did you try it with latest master?

Natim:
We don't have a build with that feature landed yet

[...]

PR #281, opened 2017-10-05
Support for Links

cedricium:
This initial commit fixes issues #253, #271, and #273. How these issues are fixed
is described below:

 - #253: When a space (or any character) is entered at index 0 of the editor
   with a link immediately following that character (as shown in the screen-grab
   for the issue), any formatting that is applied to the link will be applied
   to that character. The issue here was that the `link` format was being applied
   to space characters, but this is fixed by removing the `link` format from the
   first character of the editor.

 - #271 & #273: These issues involve the links getting any format they had
   applied to them removed as soon as the link-editing format was escaped. The
   fix for this was to capture and save the format of the link (formats like
   bold, italics, font size, etc.) then applying those same formats in addition
   to the `link` format when the url is turned into a link.

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -94,6 +94,7 @@ quill.on('text-change', function(delta) {
       endRetain += 1;
     }
     const text = quill.getText().substr(0, endRetain);
+    const format = quill.getFormat(text);
     const match = text.match(regex);
 
     if (match !== null) {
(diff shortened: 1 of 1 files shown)


Natim:
I am landing this thanks 👍

vladikoff:
@Natim found a small bug:

aa

vladikoff:
Link is styled as a link, but is not clickable

Natim:
> Link is styled as a link, but is not clickable

Also it is not a link anymore (it's content is not a link)

Natim:
The first issue is to support links that contains links, let's put this for the update of links later or remove the links feature for this release if we can't accept it.

vladikoff:
cc @ryanfeeley , see above ^^

Issue #286, opened 2017-10-06
The first letter from a link is not hyperlinked when it is pasted in a Numbered/Bulleted list


Softvision-MariusComan:
**[Affected versions]:**
- Firefox Notes 1.8.0 rc5
- Firefox 56 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest "Firefox Notes" add-on version (1.8.0 rc5) installed.
- Have a website link copied to clipboard.

**[Steps to reproduce]:**
1. Open the browser with the profiles from prerequisites.
2. Click the "Numbered List" button.
3. Paste the link from the prerequisites and observe the link.

**[Expected result]:**
- The whole link is hyperlinked.

**[Actual result]:**
- The first letter from the pasted link in not hyperlinked.

**[Notes]:**
- The issue is also reproducible if at step 2 you press the "Bulleted List" button.
- Attached a screen recording of the issue:
![hyperlink](https://user-images.githubusercontent.com/31243259/31285988-7d753024-aac5-11e7-94e6-602c55ba48e5.gif)

vladikoff:
🙀 💻

Natim:
@cedricium could you have a look at it please?

cedricium:
@Natim already have a fix in the works 👍

cedricium:
@Softvision-MariusComan can you verify this when you get a chance, should be in the latest `master` branch.

Natim:
If I understood correctly we should close it and they will validate if it is still reproducible or not :)

Softvision-MariusComan:
The issue is no longer reproducible on the latest "Firefox Notes" add-on version (1.8.0dev custom build from 10.20.2017), retested on Windows 10 x64, Ubuntu 14.04 and Mac OS 10.12 on Firefox Release (56.0) and latest Nightly (58.0a1-20171020100426).

Issue #179, opened 2017-08-02
links should be underlined and clickable.


MadhuriQA:
![2017-08-02_11h27_59](https://user-images.githubusercontent.com/26246696/28859566-b0be20cc-7775-11e7-9a85-a18bfbf6a3db.png)

- Links added in the notes should have color blue as links appears with an underline on hover. 
- It should also be clickable from the notes panel itself so that a link can be opened from the same.

johngruen:
We want to do this! adding links in the backlog, for further triage.

MadhuriQA:
@johngruen so, we will see this feature soon...am i right ?? :-)

should I create it in bugzilla as well just to have a track or follow up ?

ajomadlabs:
Can I take up this issue

Natim:
@ajomadlabs you can give it a try unless @cedricium is already working on it.

cedricium:
@Natim this should be closed, fixed via #302 (0804612)

PR #302, opened 2017-10-10
Fix for Links Issues

cedricium:
This brings in several bug fixes for the `Links` format. Specific issues fixed in this first commit are:
 - #179 - links underlined and clickable (even with other formatting options - was an issue with last PR)
 - #271 - formatted link not keeping other formatting options (like `bold`, `italics`, etc.)
 - #273 - <kbd>Space</kbd> and <kbd>Tab</kbd> characters before link are formatted as `links`
 - #286 - first letter of link not formatted as `link` when pasted into bulleted or ordered list

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -105,7 +105,7 @@ quill.on('text-change', function(delta) {
         ops.push({ retain: endRetain - url.length });
       }
 
-      let formats = {};
+      const formats = {};
       // apply any previous formatting options to the attributes object
       Object.keys(format).forEach(function(key) {
         formats[key] = format[key];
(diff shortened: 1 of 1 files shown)


Natim:
I can still reproduce #271 with this branch.

If I am typing in bold and I paste, the pasted thing is not bold
If I select a link in bold and add a bullet, it becomes not bold anymore.

Natim:
I can confirm #179, #273 and #286

cedricium:
Thanks for catching that @Natim. Looks like this branch only covers keeping preexisting formats while typing, I will work on a separate fix for pasted links in order to close #271.

Issue #332, opened 2017-10-18
Firefox Notes experiment doesn't display its author on "about:addons" page


SoftVision-CarmenFat:
[Affected versions]:
- Firefox 56.0 and up
- Firefox Notes v2.0.0a4
- Firefox Notes 1.8.0dev
- Firefox Notes Test Pilot version.

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest Firefox Notes version (v2.0.0a4 or 1.8.0dev).

[Steps to reproduce]:
1. Open the browser and navigate to "about:addons" page.
2. Click on the "Extensions" option from the left page sidebar.
3. Click on "More" option from the "Firefox Notes" experiment and observe its author.

[Expected result]:
- "By Mozilla (https://mozilla.org/)"  is displayed under the experiment title.

[Actual result]:
- No author is displayed.

[Notes]:
- The "Homepage" URL also is not displayed.
- Attached a screencast of the issue:

![notes author](https://user-images.githubusercontent.com/26000308/31713619-7f3d9b3a-b407-11e7-83b4-18eafae3d484.gif)

Softvision-CristinaBadescu:
I was able to reproduce intermittently this issue on clean new profiles using Firefox Notes 1.8.0dev (built today 2017-10-23) version, using latest Nightly build 58.0a1 (2017-10-22), on Mac 10.12. and Windows 10 x64.
Here is a screenshot of the issue, with the Web console open:
![notes](https://user-images.githubusercontent.com/31244235/31887011-673d19c4-b7ff-11e7-9dcd-7953f06f78fa.png)

@Natim, can you please take a look at the screenshot? Maybe it can help finding out why this issue is intermittent.

Also here is a screen recording with the issue, while refreshing the page:
![notes - no author](https://user-images.githubusercontent.com/31244235/31887015-6d615022-b7ff-11e7-8ef6-3c2941c10fa6.gif)

Natim:
Why are you still testing with Notes 1.8.0?
Here is the last version of master: 
[addon-1.9.0dev.zip](https://github.com/mozilla/notes/files/1414223/addon-1.9.0dev.zip)

Natim:
![screenshot](https://user-images.githubusercontent.com/229453/31994376-bca5c78a-b980-11e7-81e9-935898b5362f.png)

Softvision-CristinaBadescu:
[...]

PR #351, opened 2017-10-25
Bump version to 1.9.0dev

SoftVision-CiprianMuresan:
To avoid Firefox's autoupdate to the Test Pilot version. Fixes confusions like in #332.


Natim:
Please change it also in src/manifest.json and src/background.js

Natim:
Thanks

PR #363, opened 2017-10-27
Implement manual sync

sebastienbarbier:
On click, trigger sendMessage('load-kinto') to sync with remote instance.
Also implement a boolean isAuthenticated to disable if user requested login but is not authenticated yet.
Related to #355


sebastienbarbier:
v1 manual sync

Natim:
Thank you

Issue #330, opened 2017-10-18
The content area from "Firefox Notes" sidebar does not gain focus when it is opened or switched


Softvision-MariusComan:
**[Notes]:**
- When the Bookmarks or History sidebars are opened, their search bar is automatically focused.

**[Affected versions]:**
- Firefox Notes 2.0.0a4
- Firefox 56 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest "Firefox Notes" add-on version (2.0.0a4) installed.
- The sidebar is closed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Press alt+shift+w.
3. Write a string and observe the behavior.

**[Expected result]:**
- The string is written in the "Firefox Notes" sidebar.

**[Actual result]:**
- Nothing happens.

**[Additional Notes]:**
- This issue is also reproducible when you switch from a the Bookmarks/History sidebar to the "Firefox Notes" one.
- Attached a screen recording of the issue:
![focus notes](https://user-images.githubusercontent.com/31243259/31707501-32907972-b3f5-11e7-940d-f5bab89ab049.gif)

Softvision-MariusComan:
The issue is still reproducible on Windows 10 x64, Ubuntu 14.04 x64 and Mac 10.12, with latest Firefox release (57.0) and latest Nightly build (59.0a1) installed, using the latest master build.
It is also reproducible on 1.9.0b1 build with CKeditor

Natim:
See https://stackoverflow.com/a/8699731/186202

vladikoff:
@Natim I guess this never worked, not even in Quill. Firefox probably cannot focus on that sidebar?

vladikoff:
Why did this close ? hmm

johngruen:
Closing b/c we open into list view now

Softvision-MariusComan:
This issue is no longer reproducible with the latest version of the "Firefox Notes" add-on (4.2.0dev custom built on 07/18/2018) installed, on the latest Firefox release and Nightly. Verified on Windows 10 x64, Mac 10.13 and Ubuntu 14.04 x64.

PR #394, opened 2017-11-16
Switch to CKEditor

vladikoff:
## TODO
* [ ] ~~Missing placeholder~~ Waiting on https://github.com/ckeditor/ckeditor5/issues/479
* [x] rename 'Choose Style' to 'Aa' in the dropdown
* [ ] ~~Focus on open~~ Delayed until https://github.com/mozilla/notes/issues/330
* [x] L10N
* [x] "Hello!" is not a heading on fresh start
* [x] Missing ~~strike~~ button 
* [x] Figure out `change` events in the editor
* [x] Make welcome message use the bold `Hello!`
* [x] Bug with "Headings" when formats are mixed with lists: 
 I think we might need to fork the headings plugin or send a pull request to make that text configurable. 
* [x] Text migration
* [x] ~~Fix CSS issue with links~~ Moved to https://github.com/mozilla/notes/issues/396
* [x] Dark Theme support 
* [x] Dark Theme fixes 

* [x] Metrics based on editor usage (via @natim)
* [x] ~~RTL support~~ https://github.com/mozilla/notes/issues/414


vladikoff:
Hello @szymonkups and @Reinmar 

This part is going mostly well but could use your advice on the following:

* [ ] change text direction button + RTL support, do we need to write a plugin?
* [ ] theming the editor for a dark theme (I think we can do it by just CSSing), but would like to hear your thoughts
* [ ] do we need to write a plugin for strikethrough?

src/sidebar/panel.js
```diff
@@ -334,95 +114,26 @@ closeButton.addEventListener('click', () => {
...
-  content.forEach(node => {
-    if (node.hasOwnProperty('attributes')) {
-      Object.keys(node.attributes).forEach(key => {
-        if (styles.hasOwnProperty(key)) {
-          styles[key] = true;
-        }
-      });
-    }
-  });
+  // content.forEach(node => {
```
vladikoff: @szymonkups, @Reinmar Need a bit of help here. We used to look at node attributes in Quill to detect of features are being used. How do we detect if `bold`, `italic`, etc are used? `getData` just returns text. In Quill there is quill.getContents(); vs quill.getData();

Reinmar:
Great to see that you're moving forward with this :) I really want to see the final result.

1. The bidi feature (similar to [CKEditor 4's `bidi` plugin](https://sdk.ckeditor.com/samples/language.html)) is something that you'd need to implement. I guess that we'll implement it one day ourselves, but it's not on our roadmap yet. We could help you to start it. And as for UI's support for RTL, this is something that you could do yourself too, especially that I see that you're hacking the theme anyway :). RTLed UI is on our roadmap, but I don't know when we'll work on it.
2. See https://docs.ckeditor.com/ckeditor5/latest/framework/guides/ui/theme-customization.html. But if you can wait one month, then it'd be better because we're now switching to PostCSS which will make it easier. You can read more here: https://github.com/ckeditor/ckeditor5-design/issues/182 in the "UI improvements" section.
3. Yep, you'd need to write it, but it will be very simple. Just look at examples in https://github.com/ckeditor/ckeditor5-basic-styles/tree/master/src.

src/sidebar/panel.js
Reinmar: It'd be something like:

```js
import Range from '@ckeditor/ckeditor5-engine/src/model/range';

const range = Range.createIn( editor.document.getRoot() );

for ( const value of range ) {
    if ( value.item.is( 'text' ) ) {
        console.log( Array.from( value.item.getAttributes() ) );
    } 
}
```

Iterating over a [range](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_engine_model_range-Range.html) uses a [tree walker](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_engine_model_treewalker-TreeWalker.html) and `value` is an instance of [`TreeWalkerValue`](https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_engine_model_treewalker-TreeWalkerValue.html).

This will print which text attributes are used. You can also check which elements are used by `console.log( value.item.name )` if `value.item.is( 'element' )` is true.

Of course, this means that you need to build this code together with CKEditor. So you can't use a built version of CKEditor, but build CKEditor from source together with the code which uses it. It's explained here: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/quick-start.html.

We're now working on [making it simpler to write some basic features without building CKEditor](https://github.com/ckeditor/ckeditor5/issues/667). This means exposing more APIs somewhere in the `editor` properties.

PS. Why do you need to check which features are used in the editor?

vladikoff: > PS. Why do you need to check which features are used in the editor?

So we can learn which features are the most popular. User research, etc.. :)

vladikoff:
Migration: 

* render quill hidden if you have `notes` in storage. 
* ` quill.container.firstChild.innerHTML` get the html
* call `setData` with html. Delete the `notes` storage key
* use `notesMarkdown` for the new notes storage key
* add metrics event for migration
* only load quill if we have to do a migration
( * backup to localstorage and use that for backup maybe)

vladikoff:
@Reinmar thank you so much for the quick response. It is really helpful!

@cedricium I pushed fixes to the dark theme, it's ~70% there. Let me know if you have time to tweak the `styles-dark.css`. I won't be making changes to it for now.

cedricium:
@vladikoff sure thing, will work on the dark theme 👍

src/sidebar/panel.js
vladikoff: @Natim wanna take a look at this? 

need to uncomment the `content.forEach(node => {` function, otherwise metrics works.

vladikoff: "uncomment" => make it work. You might need to add a dependency that includes the `Range` module. 

> build CKEditor from source together with the code which uses it

vladikoff: @Natim Our build script is here: https://github.com/mozilla/notes/pull/394/files#diff-bef1bc0d0ce8255123c3de896c5a796b , builds and copies with `npm run postinstall`

Natim: It doesn't seems to work:

```
ERROR in ./src/ckeditor.js
Module not found: Error: Can't resolve '@ckeditor/ckeditor5-editor-classic/src/classiceditor' in '/home/rhubscher/mozilla/notes/ckeditor-build/src'
 @ ./src/ckeditor.js 6:0-85
```

Natim: Trying to install ``npm install --save @ckeditor/ckeditor5-editor-classic``

Natim: `npm install --save @ckeditor/ckeditor5-build-classic` nailed it.

cedricium:
@vladikoff so far I have a fix for the 'Choose heading' label overflow and I made the link color more visible for the Dark theme. Is there anything you saw that needs specific fixing?

vladikoff:
@cedricium I think mostly this menu:

image

The border is cut off and there is no highlight on the options when hovering (if compared to the normal version)

Natim:
Early Beta to start QA: [addon-1.9.0b1.zip](https://github.com/mozilla/notes/files/1487214/addon-1.9.0b1.zip)

vladikoff:
@cedricium awesome work on the dark theme 👍

vladikoff:
@Natim this should be ready for release. Could you go over the code and see if you can see anything we should adjust?

One final thing to check is the metrics. Let me know if you got time to check that before merging.

I couldn't get these to work:

* [ ] ~~Missing placeholder~~ Waiting on https://github.com/ckeditor/ckeditor5/issues/479
* [ ] ~~Focus on open~~ Delayed until https://github.com/mozilla/notes/issues/330

but I think we can ship without that. I removed the ~~strike~~ shortcut for now. Hopefully does not block us

vladikoff:
@cedricium feel free to also take a look ^

cedricium:
@vladikoff something's wrong (and I have a feeling its on my end), but I tried testing this branch and it's not working. I did:

1. `npm install` (that's all that's needed still correct, or is the command different?)
2. `npm run start`
3. Editor opens and it's completely blank (no default note shown) and the toolbar is disabled.
4. When trying to type, nothing appears in the editor. But this error shows up for every character attempted to be typed:
```node
Error: model-nodelist-insertNodes-not-node: Trying to insert an object which is not a Node instance.
Read more: https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/framework/guides/support/error-codes.html#model-nodelist-insertNodes-not-node
```

The reason I say it's me however is because I had some git mishaps earlier and tried making my local branch be the same as this upstream branch, but it may not have worked.

**Screenshots:**
screen shot 2017-11-20 at 7 28 22 pm 2
screen shot 2017-11-20 at 7 28 19 pm

vladikoff:
> -- Disregard, got it working. --

Awesome :)

Natim:
[addon-1.9.0b2.zip](https://github.com/mozilla/notes/files/1490803/addon-1.9.0b2.zip)

src/background.js
```diff
@@ -10,7 +10,7 @@ const analytics = new TestPilotGA({
   ds: 'addon',
   an: 'Notes Experiment',
   aid: 'notes@mozilla.com',
-  av: '1.9.0dev'  // XXX: Change version on release
+  av: browser.runtime.getManifest().version
```
Natim: 👍

Natim:
I am going to land this to give more time to translators to catch up before next week release.

vladikoff:
@szymonkups and @Reinmar,

this is now live: https://medium.com/firefox-test-pilot/testpilot-notes-v1-9-0-b166c6cd7353
Thanks so much for your help!!

PR #397, opened 2017-11-17
Add strike plugin.

Natim:


Natim:
@Reinmar Do you know where I can find similar SVG for the strikethrough feature? https://github.com/ckeditor/ckeditor5-basic-styles/tree/master/theme/icons

In your opinion should I rather try to contribute that directly to the ckeditor5-basic-styles?

Reinmar:
> In your opinion should I rather try to contribute that directly to the ckeditor5-basic-styles project?

If you could make a PR with the feature itself, we can add an icon for it. [We're going to create a new official theme soon anyway](https://github.com/ckeditor/ckeditor5/issues/645) so if the icon is part of ckeditor5-basic-styles we'll ensure then that it's consistent with the other ones.

Natim:
It kinda works but I have the following issues:

- [ ] The shortcut CTRL+ALT+S doesn't works
- [ ] The SVG icon doesn't show up in the toolbar.

vladikoff:
@Natim In v1.9.0 we used `CMD + SHIFT + X` for this

Natim:
> In v1.9.0 we used CMD + SHIFT + X for this

That's also @Commander advice here: https://github.com/ckeditor/ckeditor5-basic-styles/pull/58#discussion_r151722443

PR #400, opened 2017-11-20
Use P rather than 14.

Natim:
@ryanfeeley r?

p


ryanfeeley:
Sure. Let's not try to change from the editor defaults unless we are going to change them in the details too. I'm guessing that an H1 isn't just font size but breaks the line before and after.

Issue #424, opened 2017-11-21
The formatting options are greyed out if all content is deleted in a different window


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 57 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a new Firefox profile with the latest "Firefox Notes" add-on version (1.9.0b2 build with CKeditor) installed.
- Have something written in the "Notes" sidebar.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Open a new window and delete all the text from the "Notes" sidebar.
3. Return to the first opened window and observe the formatting option buttons.

[Expected result]:
- The formatting option buttons are correctly displayed and actionable.

[Actual result]:
- The formatting option buttons are greyed out and not actionable.

[Notes]:
- All opened windows are affected by this, except the one where the delete action is made.
- Here is a screen recording of the issue:
![not actionable buttons](https://user-images.githubusercontent.com/31244235/33079814-3731f92e-cedf-11e7-8d2f-3fed04e35e10.gif)

Softvision-CristinaBadescu:
This issue is no longer reproducible using latest version of Notes 1.9.0dev (custom build 2017-11-23), on latest Nightly build 59.0a1 (2017-11-22) and latest Firefox Release 57.0, on Windows 10 x64, Mac 10.12 and Ubuntu 14.04 x64.

Issue #421, opened 2017-11-21
The bullet changes into number and continues the numbered list series if there is no text between lists in a new window


SoftVision-CarmenFat:
[Affected versions]:
- Firefox 57 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (1.9.0b2 build with CKeditor) installed.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Press "Ctrl" (Cmd)+"N" keys.
3. In the new window, click the "Numbered List" button.
4. Type something, then press "Enter" key twice.
5. Click the "Bulleted List" button.
6. Type a character and observe what happens next.

[Expected result]:
- The character is correctly displayed in the bulleted list.

[Actual result]:
- The bulleted list changes into numbered list and continues the series.

[Additional Notes]:
- The issue is not reproducible if there is text between the numbered and bulleted lists.
- The issue is not reproducible with the "Firefox Notes" master build.
- Attached a screen recording of the issue:

![bulleted in numbered](https://user-images.githubusercontent.com/26000308/33074618-79b50da0-cecf-11e7-97f3-d9f5907f668f.gif)

vladikoff:
@Natim this one is pretty weird:

![numa](https://user-images.githubusercontent.com/128755/33084573-5287f4d2-ceb0-11e7-9f4a-0e35424a5b87.gif)

I will check if Markdown engine is to blame or not

Natim:
It might be because the text-change even is not triggered in case we switch list type?

vladikoff:
@Natim oooooo

vladikoff:
@Natim Ah I see, so in the markdown parser, it stripes new lines, so the linebreak does not save the state across the different windows. This is also related to this issue: https://github.com/mozilla/notes/issues/407

SoftVision-CarmenFat:
The issue is no longer reproducible on the latest version of Notes built on 11/23/2017.

Issue #420, opened 2017-11-21
The text cursor is wrongly moved at the beginning of the row if typing fast in a new window


SoftVision-CarmenFat:
[Affected versions]:
- Firefox 57 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (1.9.0b2 build with CKeditor) installed.

[Steps to reproduce]:
1. Open the browser with the profile from prerequisites.
2. Press "Ctrl" (Cmd)+"N" keys.
3. In the new window, start typing something very fast.
4. Observe the text typed.

[Expected result]:
- The text is typed correctly and the flow of the characters is constant.

[Actual result]:
- The text cursor is wrongly moved to the beginning of the first row

[Additional Notes]:
- The issue is not reproducible with the "Firefox Notes" master build.
- Attached a screen recording of the issue:

![fast typing](https://user-images.githubusercontent.com/26000308/33074512-259b1c78-cecf-11e7-89ff-41e8e0d6e3a5.gif)

Issue #407, opened 2017-11-20
The empty rows between the paragraphs are no longer displayed after reopening the sidebar


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 57 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest "Firefox Notes" add-on version (1.9.0b1 build with CKeditor) installed.
- Have at least two paragraphs with minimum an empty row between them.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Click the "Show sidebars" button from the browser toolbar to close the sidebar.
3. Click the "Show sidebars" button again and observe the sidebar's content.

**[Expected result]:**
- The content remains unchanged.

**[Actual result]:**
- The empty rows are no longer displayed.

**[Notes]:**
- The issue is also reproducible if you switch the sidebars instead of closing them.
- The issue is not reproducible with the "Firefox Notes" master build.
- Attached a screen recording of the issue:
![rec of the issue](https://user-images.githubusercontent.com/31243259/33027731-a1e91658-ce1c-11e7-9929-6493a2ba33c6.gif)

Softvision-MariusComan:
The issue is no longer reproducible on the latest version of Firefox Notes (1.9.0dev custom built on 11.23.2017) on the latest Firefox release and Nightly.

PR #427, opened 2017-11-21
Fix data races and change storage format to HTML

vladikoff:
@Natim this fixes a few bugs:

Fixes https://github.com/mozilla/notes/issues/421
Fixes https://github.com/mozilla/notes/issues/424
Fixes https://github.com/mozilla/notes/issues/420
Fixes https://github.com/mozilla/notes/issues/406
Fixes https://github.com/mozilla/notes/issues/407
Fixes https://github.com/mozilla/notes/issues/428

Related read for the reason to switch to HTML formatting: https://github.com/ckeditor/ckeditor5-markdown-gfm/issues/4

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@ckeditor/ckeditor5-essentials": "^1.0.0-alpha.2",
     "@ckeditor/ckeditor5-heading": "vladikoff/ckeditor5-heading#short-heading",
     "@ckeditor/ckeditor5-list": "^1.0.0-alpha.2",
-    "@ckeditor/ckeditor5-markdown-gfm": "vladikoff/ckeditor5-markdown-gfm#data-as-html",
+    "@ckeditor/ckeditor5-markdown-gfm": "1.0.0-alpha.2",
     "@ckeditor/ckeditor5-paragraph": "^1.0.0-alpha.2",
     "babel-minify-webpack-plugin": "^0.2.0",
     "css-loader": "^0.28.7",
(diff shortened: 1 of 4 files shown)


vladikoff:
@Natim here's the only change I made in the gfm processor: https://github.com/vladikoff/ckeditor5-markdown-gfm/commit/407840c309235750370868c511e6390a50621e49

I feel like storing the data as HTML is the way to go, especially if we plan to have COLOR and RTL support in the future.

vladikoff:
@Natim I feel like we should include this in the release because the multi window data race is pretty bad

vladikoff:
Ah changing the Heading does not fire `text-change`

vladikoff:
@Natim Ok should be good to go 👍 , test it out

package.json
```diff
@@ -17,7 +17,7 @@
     "@ckeditor/ckeditor5-essentials": "^1.0.0-alpha.2",
     "@ckeditor/ckeditor5-heading": "vladikoff/ckeditor5-heading#short-heading",
     "@ckeditor/ckeditor5-list": "^1.0.0-alpha.2",
-    "@ckeditor/ckeditor5-markdown-gfm": "^1.0.0-alpha.2",
+    "@ckeditor/ckeditor5-markdown-gfm": "vladikoff/ckeditor5-markdown-gfm#data-as-html",
```
Natim: Can we make this part of this repository?

vladikoff: i will try!!

vladikoff: Fixed by a function override instead of using the fork

Natim: So shall we get back to `^1.0.0-alpha.2`?

PR #445, opened 2017-11-23
Add new Sync strings

vladikoff:


locales/en-US/notes.properties
```diff
@@ -23,6 +32,11 @@ disableSync=Disable Sync
 # structure doesn't work for your locale, you can translate this as "Last sync:
 # {date}".
 syncComplete=Synced {date}
+syncCompleteAt=Synced at {date}
```
flodolo: Is `{date}` a date or a time?

Natim: It is the same as for Saved at (basically a time)

Natim:
```
# LOCALIZATION NOTE (savedComplete): {date} is the time of saving. If this
# structure doesn't work for your locale, you can translate this as "Saved (
# {date})".
savedComplete=Saved at {date}
```

flodolo: Can you give me an example? Is it just the time, or date+time?

vladikoff: Sorry for adding this confusion, we already had one string like this and I reused the note. @ryanfeeley is gonna add a mockup of that this should look like below

ryanfeeley: @flodolo Because it should not be possible to see a timestamp from a previous day (as it syncs on open) we can just show the time. Updated the mockup. https://mozilla.invisionapp.com/d/main#/console/11122662/235284916/preview

flodolo: So, if I'm understanding this correctly: in most cases it will be a time, but it still could be a date if I open Firefox after a few days. If that's the case, how does that work for English?

vladikoff: I feel like it is not `{date}` anymore, but it is `{time}` ? @flodolo should we change the string?

image

flodolo: If you're sure that's always only a time (I tried to figure it out the other day on Slack), probably yes.

vladikoff: @flodolo I think we are gonna go with `{time}` we cannot fit the full date in there. cc @ryanfeeley

vladikoff: @flodolo should i do `syncCompleteAt2` and `syncComplete2` to signal that we changed these ?

flodolo: That, or syncCompleteTime

vladikoff: Thanks so much!

PR #511, opened 2017-12-13
Preparing 2.1.0 release

vladikoff:


Natim:
I am not mad, but I feel that in the future, we should try to avoid releasing a version without review, without waiting for the tests to run and without following SemVer.

SoftVision-PaulOiegas:
Totally agree with @Natim. Also this a very bad timing for us since half of the team in in Austin and we have a lot of work on our plate. We won't be able to test Notes for another day or even next week only...

Natim:
@SoftVision-PaulOiegas don't worry about testing the features for this version, there were no code change as you can see here: https://github.com/mozilla/notes/compare/2.0.0...2.1.0

They did the release only to trigger the fix in the deployment procedure.

SoftVision-PaulOiegas:
Yeh, just looked over the commit and noticed that too. My bad.

Issue #519, opened 2017-12-19
The "Firefox Notes Feedback" page is wrongly opened if you click the area around the "Notes" menu button


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 57.0.2 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest "Firefox Notes" add-on version (2.2.0dev custom build from 12.19.2017) installed.
- You are already logged into "Firefox Notes".

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Click anywhere near the "Notes" (three dots) menu button.
3. Observe the behavior.

**[Expected result]:**
- Nothing happens.

**[Actual result]:**
- The "Firefox Notes Feedback" page is opened. 

**[Notes]:**
- The issue is also reproducible with the TxP version of "Firefox Notes".
- The issue is reproducible only if you are logged into "Firefox Notes".
- Attached a screen recording of the issue:
![rec of the issue](https://user-images.githubusercontent.com/31243259/34151566-519e7cde-e4b4-11e7-959b-58bca269ed45.gif)

himanish-star:
I would like to try this

vladikoff:
@himanish-star please!

Softvision-MariusComan:
The issue is no longer reproducible with the latest version of the "Firefox Notes" add-on (2.2.0dev custom build from 01.12.2018) installed, on the latest Firefox release and Nightly.

PR #548, opened 2018-01-05
The Firefox Notes feedback page is no longer wrongly opened on clicking the area around the 'notes' menu button

himanish-star:
fixes: #519 

I have fixed this but the only problem is that the size of the button background has been increased, **which can only be seen when we hover over the menu button**. Well according to me all things work fine, however if you don't prefer these changes then I will have to approach this some other way.

Here is a GIF of the changes
simplescreenrecorder-2018-01-05_15 56 58

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -213,6 +213,7 @@ function reconnectSync () {
 function disconnectFromSync () {
   waitingToReconnect = false;
   disconnectSync.style.display = 'none';
+  giveFeedbackButton.style.display = 'inherit';
   isAuthenticated = false;
   setAnimation(false, false, false); // animateSyncIcon, syncingLayout, warning
   setTimeout(() => {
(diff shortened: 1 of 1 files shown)


shikhar-scs:
Good work @himanish-star :+1:

Natim:
Let's see what @ryanfeeley think about that :)

ryanfeeley:
Good eye @himanish-star! I think the hover was never added, but I did include it in the original design. I try to match the hover style of the formatting buttons (and other toolbar buttons in Firefox).. Rounded corner, centered vertically in space. https://mozilla.invisionapp.com/share/6VBUYHMRB#/235284916_Desktop_Sidebar

himanish-star:
Ok I'm on it

himanish-star:
I have made the changes. This issue was being caused due to overlapping of buttons and thus when we were clicking the side areas of the menu button the feedback page was getting opened. For **illustration purpose** only I have chosen to show those buttons in **blue**.
Here's a GIF

ezgif com-video-to-gif 6

src/sidebar/panel.js
```diff
@@ -71,6 +71,7 @@ ClassicEditor.create(document.querySelector('#editor'), {
       });
 
       enableSync.onclick = () => {
+        giveFeedbackButton.style.display = 'none';
```
Natim: Can you put that inside the enableSyncAction function?

himanish-star: Sure as you wish

himanish-star:
There is still one problem which I'm trying to fix. I'll update this PR soon after fixing it.

Natim:
Good job @himanish-star thanks !

Issue #526, opened 2017-12-19
The "Sync Your Notes" button is slightly moving when it's being clicked


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 57.0.2 and up

[Affected Platforms]:
- All Windows
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (2.2.0dev custom build from 12.19.2017) installed.
- You are already logged into "Firefox Notes" account.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites.
2. Click "Sync Your Notes" button and observe the behavior.

[Expected result]:
- Sync button doesn't move and the animation begins.

[Actual result]:
- Sync button is slightly moving before the animation begins.

[Notes]:
- This issue is also reproducible when you try to reconnect to sync or when you try to log in.
- Here is a screen recording of the issue:
![moving sync up button](https://user-images.githubusercontent.com/31244235/34165168-339cc058-e4e4-11e7-8896-24391f1f2178.gif)

himanish-star:
I would surely like to look into this.

Softvision-CristinaBadescu:
This issue is no longer reproducible using "Firefox Notes" add-on v2.2.0dev custom build from 01.10.2018, on latest Nightly build 59.0a1 (2018-01-10), on Windows 10x64 and Ubuntu 14.04 x64.

Softvision-CristinaBadescu:
This issue is still reproducible using Firefox Notes v3.0.0dev, on latest Nightly build 60.0a1 (2018-02-20), on Windows 10 x64 and Arch Linux 4.12. Considering this, I'm going to reopen the issue.

vladikoff:
Should be fixed via 

https://github.com/mozilla/notes/pull/756
https://github.com/mozilla/notes/pull/752

Let us know! It doesn't jump for me in OS X anymore...

Softvision-CristinaBadescu:
As I've mentioned in the affected platforms, only on Windows and Linux OS'es this issue is reproducible. Indeed, on Mac OS'es this behavior does not occur.

I'm going to reopen this, since it's still reproducible using Firefox Notes v3.1.0dev (custom built on 2018-03-12), on latest Nightly build 60.0a1 (2018-03-11), on Windows 10 x64 and Linux 4.12.

vladikoff:
Merged a fix for it, will be part of 3.1 release so we can test 👍

[...]

PR #551, opened 2018-01-06
The sync button no longer dislocates on being clicked

himanish-star:
fixes : #526 

Changes can be previewed below

ezgif com-video-to-gif 3

### DIFF
--- a/src/sidebar/panel.html
+++ b/src/sidebar/panel.html
@@ -31,11 +31,11 @@
     <div id="footer-buttons" class="savingLayout">
       <div>
         <span id="saving-indicator"></span>
-        <button id="enable-sync" class="notsyncing">
+        <span id="enable-sync" class="notsyncing">
           <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 16 16">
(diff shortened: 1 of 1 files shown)


himanish-star:
@Natim , Could you have a look into this. As the CI test which failed has nothing to do with my commit as I have only made a minor change

Natim:
Is it normal that it doesn't turn into a Syncing change message?

himanish-star:
>Is it normal that it doesn't turn into a Syncing change message?

Oh sorry, didn't notice that this got changed too. I will correct this.

Natim:
@himanish-star can you try to keep button and add a display: inline for the button in the CSS?

himanish-star:
@Natim , currently I'm facing some issues with my browser. When I click the sync button the login page opens up and then when I try to log in I get this
screenshot from 2018-01-08 19-43-44
I'm really trying to find out why this is happening. I was playing around with the settings today for one of the projects named pdf.js to try an install a plugin and from then on my sync is not working. Please excuse me for a while as I'm trying to resurrect back things. I will soon make the channges

Natim:
You can probably use a new profile for your tests? Or use `npm run start`?

himanish-star:
> @himanish-star can you try to keep button and add a display: inline for the button in the CSS?

Yeah sure doing that now.

> Is it normal that it doesn't turn into a Syncing change message?

Well on clicking the sync button previously before the changes also nothing of this sort used to happen. As I recently tested the master branch and found out that it doesn't turn into a Syncing change message.

himanish-star:
>@himanish-star can you try to keep button and add a display: inline for the button in the CSS?

@Natim , nothing seems to work. Only if use `<span>` instead of `<button>` things become correct. The display inline doesn't fix this as I tried doing so and it didn't work

Issue #487, opened 2017-12-06
Allow pressing "Reconnect to Sync" as a button


vladikoff:
Currently you have to press the Sync arrows to reconnect

utkarsh858:
I would like to work on it

vladikoff:
Ok!

utkarsh858:
Thanks

PR #558, opened 2018-01-09
'Reconnect to Sync' text field acts as a button #487

utkarsh858:
fixes : #487 

The text field will work as same as the Sync Button.

### DIFF
--- a/src/sidebar/menu.css
+++ b/src/sidebar/menu.css
@@ -1,4 +1,4 @@
-/**
+#/**
  * Copyright 2015 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/sidebar/menu.css
+++ b/src/sidebar/menu.css
@@ -1,4 +1,4 @@
-#/**
+/**
  * Copyright 2015 Google Inc. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


utkarsh858:
@vladikoff  Can you look into it .

src/sidebar/menu.css
```diff
@@ -1,4 +1,4 @@
-/**
+#/**
```
Natim: I believe this should not be present here.

src/sidebar/panel.js
```diff
@@ -69,11 +69,16 @@ ClassicEditor.create(document.querySelector('#editor'), {
         }
         ignoreNextLoadEvent = false;
       });
-
+      
+      savingIndicator.onclick = () => {
+        enableSyncAction(editor);
```
Natim: I believe this should be deactivated during saving (as for enableSync)

Natim: Nevermind it seems to be handled in the enableSyncAction itself: https://github.com/utkarsh858/notes/blob/6135f7841191671facbfcefd12f91d5f033c9e46/src/sidebar/panel.js#L238

src/sidebar/panel.js
```diff
@@ -205,6 +210,7 @@ function reconnectSync () {
   isAuthenticated = false;
   setAnimation(false, true, true); // animateSyncIcon, syncingLayout, warning
   savingIndicator.textContent = browser.i18n.getMessage('reconnectSync');
+  savingIndicator.style="cursor:pointer;";
```
Natim: If it works for all sidebar actions, why put the cursor only for reconnect to sync?

utkarsh858: Because the task for that text field only.

PR #559, opened 2018-01-09
Add menu item to open notes in a tab rather than in the sidebar.

### LINKED ISSUE #470: Provide a button to open notes in a Tab
The hardest part here is: we might not be able to hide the sidebar on our own, so we get 2 editors

Natim:
Fixes #470 

The issue with this PR is that we only access the menu after logging in to sync...
Also we can still open in a new tab while being in a tab already.

### DIFF
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -54,3 +54,5 @@ textDirectionTitle=Text Direction
 themeLegend=Theme
 defaultThemeTitle=Default
 darkThemeTitle=Dark
+
+openInTab=Open in a tab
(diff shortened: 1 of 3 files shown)


ryanfeeley:
@Natim @vladikoff  Hmm... where did this idea come from? I need more context to evaluate this idea, and consider the complexities doing so will introduce.

Natim:
The idea comes from @vladikoff :) https://github.com/mozilla/notes/issues/470

Natim:
Maybe we could add the link in the preference panel?

vladikoff:
@Natim , okay, @ryanfeeley said let's delay this until we have the meatball menu in react

ryanfeeley:
image

grahamperrin:
Oh, with Firefox Quantum limited to a single sidebar, this was one of the first features I sought :-(

Issue #562, opened 2018-01-09
The sidebar background is not highlighted when dragging text over it, if Dark theme is enabled


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 57.0.4 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (2.2.0dev custom build from 01.09.2018) installed.
- Have the "Dark" theme selected for the "Firefox Notes" add-on from about:addons->Extensions tab->Notes.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Navigate to any website and select a text from the page.
3. Drag the text over the sidebar and observe the sidebar's background.

**[Expected result]:**
- The background is highlighted.

**[Actual result]:**
- Only the background's border is highlighted..

**[Notes]:**
- The issue is not reproducible with the TxP version of "Firefox Notes".
- Attached a screen recording of the issue:
![rec of the issue 1](https://user-images.githubusercontent.com/31243259/34726548-0211d0ee-f55d-11e7-9ba4-177a8b5e23c0.gif)

Natim:
@himanish-star can I let you look into that bug?

himanish-star:
@Natim , no doubt about it. It's my pleasure.

PR #565, opened 2018-01-10
The sidebar now gets focused on dragging and dropping text when dark theme is enabled

himanish-star:
fixes : #562 
Below is a GIF of the changes made
ezgif com-video-to-gif 8
~I have made a spelling mistake in my commit message. Shall I correct it ?~

### DIFF
--- a/src/sidebar/editor.js
+++ b/src/sidebar/editor.js
@@ -1,5 +1,7 @@
 /* exported customizeEditor, getPadStats, localizeEditorButtons, setAnimation, formatFooterTime */
 function customizeEditor(editor) {
+  const mainEditor = document.querySelector('.ck-editor__main');
+
   // Disable right clicks
   // Refs: https://stackoverflow.com/a/737043/186202
   document.querySelectorAll('.ck-toolbar, #footer-buttons').forEach((sel) => {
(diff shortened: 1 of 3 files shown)


src/sidebar/editor.js
```diff
@@ -17,22 +17,49 @@ function customizeEditor(editor) {
   });
   
   document.addEventListener('dragover', () => {
-    const mainEditor = document.querySelector('.ck-editor__main');
-    mainEditor.style.border = '1px solid green';
-    mainEditor.style.backgroundColor = '#dcd6da';
+    const theme = browser.storage.local.get(['theme']);
+    theme.then(function getTheme(data) {
+      if (data.theme === 'default') {
```
Natim: I would recommend doing that with a class that we can define in both styles.css and styles-dark.css rather than having this kind of if around the theme here.

himanish-star: @Natim , do you mean that instead of setting colors here , I have to just add a class and then add the class and remove the class upon requirement right ?

Natim: Yes

himanish-star: Yo got it

himanish-star: Have a look now

src/sidebar/editor.js
```diff
@@ -17,22 +17,43 @@ function customizeEditor(editor) {
...
   
   document.addEventListener('dragover', () => {
+    const theme = browser.storage.local.get(['theme']);
     const mainEditor = document.querySelector('.ck-editor__main');
-    mainEditor.style.border = '1px solid green';
-    mainEditor.style.backgroundColor = '#dcd6da';
+    const editableEditor = document.querySelector('.ck-editor__editable');
+    theme.then(function getTheme(data) {
+      if (data.theme === 'default') {
+        mainEditor.classList.add('drag-n-drop-focus');
```
Natim: No need for a if anymore. (the all point of the refactoring you did)

himanish-star: Actually the thing is that as you can see I have used `mainEditor` and `editableEditor`. For the dark theme simply using `mainEditor` doesn't work as the below line gives `.ck-editor__editable` an extra background color which doesn't allow the changes in `mainEditor` to show.
https://github.com/mozilla/notes/blob/102c8749351311eeb29b5e68b54043a45b57a425/src/sidebar/styles-dark.css#L94
If you want I can remove this background then there would be no need of `if`

Natim: Yes but you should be able to use css inheritance to merge them nevertheless. I can give it a try if you want.

himanish-star: Ok, I'll try this now. If it doesn't work , then for sure you can take over

himanish-star: How about now ?

Natim:
Thanks @himanish-star good job !

himanish-star:
Thanks for merging.

Issue #569, opened 2018-01-10
Increase line height of list items


ryanfeeley:
<img width="362" alt="image" src="https://user-images.githubusercontent.com/68213/34797096-5aa8039a-f625-11e7-9088-0e4f6b9c6a5e.png">

PR #579, opened 2018-01-15
Increase line height of list items

cedricium:
Fix for #569. How does the change look @ryanfeeley?

**Before:**
screen shot 2018-01-15 at 1 42 13 am

**After:**
screen shot 2018-01-15 at 1 41 40 am

### DIFF
--- a/src/sidebar/styles.css
+++ b/src/sidebar/styles.css
@@ -50,6 +50,10 @@ body .ck-editor .ck-editor__editable {
   padding-left: 20px;
 }
 
+.ck-editor__editable ol > li, .ck-editor__editable ul > li {
+  line-height: 1.5rem;
+}
+
 /**
 Custom list types for indents
 https://github.com/mozilla/notes/issues/413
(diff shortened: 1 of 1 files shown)


src/sidebar/styles.css
```diff
@@ -50,6 +50,10 @@ body .ck-editor .ck-editor__editable {
   padding-left: 20px;
 }
 
+.ck-editor__editable ol > li, .ck-editor__editable ul > li {
+  line-height: 1.5rem;
```
Natim: Is 1rem colliding?

cedricium: @Natim it is. 1rem is 16px which I think matches the current P font-size, thus the reason for the overlapping.

Natim: Ok thanks !

ryanfeeley:
@cedricium Looks good, although my example was using code formatting.

- `I wonder`
- `how GitHub’s`
- `text editor`
- `does line-height`
- ` as code`
- ok, good I'm not crazy
- i think they have a nice line height

Issue #583, opened 2018-01-16
The "Firefox Notes" toolbar button is not visible when the browser's dark theme is activated


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 57.0.4 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (2.2.0dev custom build from 01.16.2018) installed.
- Have the "Dark" theme enabled from about:addons->Themes.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Observe the "Firefox Notes" toolbar button.

**[Expected result]:**
- The button is visible.

**[Actual result]:**
- The button blends in with the background.

**[Notes]:**
- The issue is not reproducible with the TxP version of "Firefox Notes".
- Attached a screen recording of the issue:
![rec of the issue](https://user-images.githubusercontent.com/31243259/34995398-08be0c64-fadf-11e7-951b-5ed58a156dbe.gif)

Natim:
When we merged it it was looking like that:

https://user-images.githubusercontent.com/3018618/34559218-57dc0ae6-f141-11e7-9482-5a0e107cf3c4.gif

Natim:
I can reproduce on Linux but we can't on Mac (we tried with release/beta and nightly versions) I am suspecting a Firefox bug.

Natim:
We can reproduce on Windows.

Natim:
It seems to work only on retina displays...

sebastienbarbier:
Could solve this issue by adding in manifest "theme_icons" size 16 in the list. Will push a fix soon.

Natim:
Released in https://github.com/mozilla/notes/files/1642619/notes-2.1.1rc2.zip

Softvision-MariusComan:
The issue is no longer reproducible with the latest version of the "Firefox Notes" add-on (2.2.0dev custom build from 01.25.2018) installed, on the latest Firefox release and Nightly.

PR #590, opened 2018-01-18
Add in manifest browser_action theme_icons size 16

sebastienbarbier:
Should fix issue #583


src/manifest.json
```diff
@@ -32,6 +32,10 @@
     "theme_icons": [{
       "light": "icons/notes-48-light.png",
       "dark": "icons/notes-48-dark.png",
+      "size": 16
+    }{
```
Natim: missing coma ?

Issue #597, opened 2018-01-18
Use syncing instead of saving while logged in


Natim:

Natim:
## STEPS TO REPRODUCE

    Enable Syncing of Notes
    Edit a note

## EXPECTED RESULTS

    After a note has been edited, footer should say "Syncing…"
    Before a note has been edited it shoudl say "Last Synced HH:MM"

## ACTUAL RESULTS

    It does this but also has Save and Saving…

himanish-star:
@Natim , I would like to fix this issue. May I ?

Natim:
Yeah sure go ahead.

Issue #532, opened 2017-12-21
Text isn't dropped in the text caret indicated position


Softvision-CristinaBadescu:
[Notes]:
- Also, you cannot drop text from "Notes" sidebar into another position in "Notes" sidebar.

[Affected versions]:
- Firefox 57.0.2 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (2.2.0dev custom build from 12.21.2017) installed.
- Have something written in the "Notes" sidebar.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites.
2. Navigate to a website and select a text (eg. https://en.wikipedia.org/wiki/Tree).
3. Drag the text in a desired position in the "Notes" sidebar.
4. Drop it and observe where the text is placed.

[Expected result]:
- The text is placed in the indicated position by the text caret.

[Actual result]:
- The text isn't placed in the indicated position by the text caret, it's wrongly placed in the text caret last position.

[Additional notes]:
- This issue is not reproducible with other text editors, such as Google Docs or LibreOffice.
- Here is a screen recording of the issue:
![drop text in text caret indicated position](https://user-images.githubusercontent.com/31244235/34255503-af65b9bc-e659-11e7-8d77-e965aa1739bc.gif)

himanish-star:
@Softvision-CristinaBadescu , I would like to take up this issue. May I  ?

Softvision-CristinaBadescu:
This isn't up to me, but @Natim or @vladikoff  are entitled to say if you can start working on this. Thanks for understanding :)

himanish-star:
@Natim, @vladikoff. I would like to take this issue. Also I want to know that how do we get the current cursor position in the CKeditor. I tried using [`getRanges`](https://docs.ckeditor.com/ckeditor5/latest/api/module_engine_view_selection-Selection.html#getRanges) but it didn't work. Could anyone of you both guide me here

vladikoff:
@himanish-star first check if the same bug is present on https://ckeditor5.github.io/

vladikoff:
@himanish-star we need to confirm if it is a CKEditor bug or something else

PR #467, opened 2017-11-30
Update Sync strings and add better L10N descriptions

vladikoff:
@flodolo r?

PR #601, opened 2018-01-21
syncing used instead of saving while logged in

himanish-star:
fixes : #597 

Changes can be seen in the GIF below
ezgif com-crop

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -122,7 +122,9 @@ ClassicEditor.create(document.querySelector('#editor'), {
               setAnimation(true); // animateSyncIcon, syncingLayout, warning
               syncingInProcess = true;
             }
-            if (! waitingToReconnect) {
+            if (! waitingToReconnect && isAuthenticated) {
+              savingIndicator.textContent = browser.i18n.getMessage('syncProgress');
+            } else if (! waitingToReconnect) {
               savingIndicator.textContent = browser.i18n.getMessage('savingChanges');
             }
             // Disable sync-action
@@ -139,7 +141,7 @@ ClassicEditor.create(document.querySelector('#editor'), {
             syncingInProcess = false;
             break;
           case 'text-saved':
-            if (! waitingToReconnect) {
+            if (! waitingToReconnect && ! isAuthenticated) {
               // persist reconnect warning, do not override with the 'saved at'
               savingIndicator.textContent = browser.i18n.getMessage('savedComplete2', formatFooterTime());
             }
@@ -285,7 +287,7 @@ function getLastSyncedTime() {
     isAuthenticated = true;
     setAnimation(false, true);
   } else {
-    savingIndicator.textContent = browser.i18n.getMessage('savedComplete2', formatFooterTime());
+    savingIndicator.textContent = browser.i18n.getMessage('changesSaved', formatFooterTime());
   }
 }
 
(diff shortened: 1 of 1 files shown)


locales/en-US/notes.properties
```diff
@@ -17,7 +17,7 @@ disableSync=Disable Sync
 reconnectSync=Reconnect to Sync
 syncError=Sync Error
 
-savingChanges=Saving changes…
+savingChanges=Syncing…
```
flodolo: If you change a string you also need to change the associated ID
https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings

Also make sure to update the reference in comments (if available)

Natim:
This is not the proper way of fixing the issue. You can't change the property file like that.

himanish-star:
@Natim , I was about fix this. Could you reopen this PR ? I'll try this another way.

himanish-star:
This is with respect to https://github.com/mozilla/notes/issues/597#issuecomment-358702806

> After a note has been edited, footer should say "Syncing…"

I tried searching for the word "Syncing" but I found out "Syncing changes". Thus I have used this as you have told me that I can't edit the property file like that.
> Before a note has been edited it should say "Last Synced HH:MM"

Same for this I couldn't find "Last Synced" so I did it with "Synced at" instead.
> It does this but also has Save and Saving…

So do you want me to remove this. For now I have removed this , if you want it back then I'll fix this back too :smiley:
Here's the new GIF
simplescreenrecorder-2018-01-22_16 56 02
__________________________
If this is not what you want, then could you explain me a bit more :smiley:

himanish-star:
@Natim , any views on the recent changes ?

Natim:
If we need to go back to _Synced 7:24 PM_ it is a regression of https://github.com/mozilla/notes/pull/467 and @ryanfeeley @vladikoff and @flodolo need to talk to figure out what made that change that I don't remember.

Natim:
Writting the following made me understand why we added the at inside:

```
# LOCALIZATION NOTE (syncComplete): {time} is the time of last sync. If this
# structure doesn't work for your locale, you can translate this as "Synced at {time}".
syncComplete3=Synced {time}
```

@ryanfeeley is that really what we want?

ryanfeeley:
@Natim Yes, an example like *Synced 7:24 PM* is perfect. Same for *Saved 7:24 PM*.

himanish-star:
ezgif com-video-to-gif
This is a new GIF. Please see the changes

ryanfeeley:
@himanish-star Is it putting the new text at the end (where the cursor is) or the beginning? Hard to tell as it's the same text.

himanish-star:
Well it's putting the text according to the previous caret position. There is an issue already on this #532

src/sidebar/panel.js
```diff
@@ -122,7 +122,9 @@ ClassicEditor.create(document.querySelector('#editor'), {
               setAnimation(true); // animateSyncIcon, syncingLayout, warning
               syncingInProcess = true;
             }
-            if (! waitingToReconnect) {
+            if (! waitingToReconnect && isAuthenticated) {
```
Natim: I guess you can have a if for the `!waitingToReconnect` and another one for the `isAuthenticated` condition.

himanish-star: Oh, I just saw this now. Well thanks for changing :smiley:

PR #609, opened 2018-01-24
Use kinto-load at startup to be much quicker.

### LINKED ISSUE #538: Start time should not be about 3 seconds
When sync is enabled, start time is about 3 seconds: 

    Start time:  2851

We should aim at making it a couple hundreds milliseconds.

Natim:
Fixes #538


himanish-star:
I really had no idea that this was the change. Seriously, I had spent quite a lot of time thinking about what the change might be.

PR #610, opened 2018-01-24
Display a message when the notepad is growing too much.

### LINKED ISSUE #546: Add a limit to the number of char in the editor to avoid it breaking after 9000 lines
Maybe we should port: https://ckeditor.com/old/forums/CKEditor-3.x/Character-Count-Plugin

Natim:
Fixes #546

### DIFF
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -19,7 +19,7 @@ syncError=Sync Error
 
 savingChanges=Saving changes…
 changesSaved=All changes saved
-maximumPadSizeExceeded=The multiple notes feature is coming soon. Meanwhile your note reached its maximum size, please cleanup before adding more content to it.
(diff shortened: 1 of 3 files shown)


locales/en-US/notes.properties
```diff
@@ -19,7 +19,7 @@ syncError=Sync Error
 
 savingChanges=Saving changes…
 changesSaved=All changes saved
-
+maximumPadSizeExceeded=The multiple notes feature is coming soon. Meanwhile your note reached its maximum size, please cleanup before adding more content to it.
```
flodolo: I think the verb in English is "clean up", while "cleanup" is the noun? Also not sure if there should be a comma before "your note".

Natim:
I filed #618 about the colors.

Issue #631, opened 2018-01-30
The "Firefox Notes Feedback" page is not opened in a new tab


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 58.0 and up

**[Affected Platforms]:**
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (2.2.0rc2) installed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Click the "Give Feedback" button from the bottom right part of the sidebar.
3. Observe the behavior.

**[Expected result]:**
- The "Firefox Notes Feedback" page is opened in a new tab.

**[Actual result]:**
- The "Firefox Notes Feedback" page is opened in the same tab.

**[Additional Notes]:**
- The issue is reproducible with the TxP version of "Firefox Notes".
- Attached a screen recording of the issue:
![rec-of-the-issue](https://user-images.githubusercontent.com/31243259/35575393-ee3917ce-05e4-11e8-975a-ff548f750036.gif)

apoorvaagupta:
Hello, May I take up this issue if no one else is working on it?

Natim:
Yes please go ahead @apoorvaagupta

PR #635, opened 2018-01-30
Open feedback form in a new tab

apoorvaagupta:
solves #631

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -28,23 +28,19 @@ enableSync.setAttribute('title', browser.i18n.getMessage('syncNotes'));
 giveFeedbackButton.setAttribute('title', browser.i18n.getMessage('feedback'));
 giveFeedbackMenuItem.text = browser.i18n.getMessage('feedback');
 
+function giveFeedbackCallback(e) {
+  e.preventDefault();
+  browser.tabs.create({
+    url: SURVEY_PATH
+  });
+}
+
 giveFeedbackButton.addEventListener('dragstart', (e) => {
   e.preventDefault();
 });
 
-giveFeedbackButton.addEventListener('click', (e) => {
-  e.preventDefault();
-  browser.tabs.create({
-    url: SURVEY_PATH
-  });
-});
-
-giveFeedbackMenuItem.addEventListener('click', (e) => {
-  e.preventDefault();
-  browser.tabs.create({
-    url: SURVEY_PATH
-  });
-});
+giveFeedbackButton.addEventListener('click', giveFeedbackCallback);
+giveFeedbackMenuItem.addEventListener('click', giveFeedbackCallback);
 
 // ignoreNextLoadEvent is used to make sure the update does not trigger on other sidebars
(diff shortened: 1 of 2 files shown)


src/sidebar/panel.js
```diff
@@ -28,7 +28,9 @@ enableSync.setAttribute('title', browser.i18n.getMessage('syncNotes'));
 giveFeedbackButton.setAttribute('title', browser.i18n.getMessage('feedback'));
 giveFeedbackMenuItem.text = browser.i18n.getMessage('feedback');
 giveFeedbackButton.setAttribute('href', SURVEY_PATH);
+giveFeedbackButton.setAttribute('target', '_blank');
```
vladikoff: @apoorvaagupta could you please also add `'rel', 'noopener'` besides this new `target` attribute

See https://html.spec.whatwg.org/multipage/links.html#link-type-noopener

src/sidebar/panel.js
```diff
@@ -28,7 +28,11 @@ enableSync.setAttribute('title', browser.i18n.getMessage('syncNotes'));
...
+giveFeedbackButton.setAttribute('rel', 'noopener');
 giveFeedbackMenuItem.setAttribute('href', SURVEY_PATH);
+giveFeedbackMenuItem.setAttribute('target', '_blank');
```
vladikoff: @apoorvaagupta I tried this locally and instead of opening a new tab this new code opens a new window. See the original issue, they expect a new tab instead. Not sure if we can make that happen. 

@Softvision-MariusComan thoughts?

apoorvaagupta: Hey!
@vladikoff The value '_blank' of the attribute 'target' opens the link in a new browsing context. It may be a tab(usually) or a new window depending on the browser settings. This setting cant be customised as it is upto the user.

apoorvaagupta: The explanation can be viewed under the target section of the following link:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

Natim: I think you can use the webextension tabs.open API to make sure to always create a tab.

apoorvaagupta: Thankyou @Natim I will just check that and try to work on it. :)

apoorvaagupta:
@vladikoff @Natim Please review!

src/sidebar/panel.js
```diff
@@ -27,13 +27,25 @@ let syncingInProcess = false;
...
 giveFeedbackMenuItem.text = browser.i18n.getMessage('feedback');
-giveFeedbackButton.setAttribute('href', SURVEY_PATH);
-giveFeedbackMenuItem.setAttribute('href', SURVEY_PATH);
 
 giveFeedbackButton.addEventListener('dragstart', (e) => {
   e.preventDefault();
 });
 
+giveFeedbackButton.addEventListener('click', (e) => {
+  e.preventDefault();
```
vladikoff: You are repeating the same function code for both of these. You should create a new function and use it for both `addEventListener`s

Issue #629, opened 2018-01-30
When hovering the "Synced at.." text, the Sync email is no longer displayed after switching sidebars


Softvision-CristinaBadescu:
[Affected versions]:
- Firefox 58.0 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (2.2.0dev custom build from 2018-01-30) installed.
- Be logged in a Firefox Notes account.

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites. 
2. Switch the "Notes" sidebar to another one, eg. "History".
3. Switch back to the "Notes" sidebar.
4. Hover over the "Synced at.." text and observe the behavior.

[Expected result]:
- A tooltip with the Sync email is displayed.

[Actual result]:
- No tooltip is displayed.

[Notes]:
- This issue is also reproducible if you restart the browser, without switching the sidebars.
- This issue is reproducible with Firefox Notes from Test Pilot experiments [page](https://github.com/mozilla/notes/issues).
- Here is a screen recording of the issue:
![display the email address](https://user-images.githubusercontent.com/31244235/35562630-a2065574-05bc-11e8-8e36-e2b4f9e6e736.gif)

himanish-star:
@Natim, shall I try to fix this ?

vladikoff:
@himanish-star sure, I think you need to cache the email somewhere after the user authenticates

himanish-star:
Sure, I'll start looking into this

vladikoff:
Should be fixed via https://github.com/mozilla/notes/pull/637

Softvision-CristinaBadescu:
This issue is no longer reproducible using latest Firefox Notes version 2.2.0 dev (custom built 2018-02-01), on latest Nightly build 60.0a1 (2018-01-31), on Windows 10 x64, Mac 10.12.6 and Arch Linux 4.12.

However, I've noticed that the text from the tooltip is different: before it displayed only the Sync email address, and now it displays "Sync with [e-mail address]", as you can see in the below screenshot. Should I log a new issue for this?
![sync with](https://user-images.githubusercontent.com/31244235/35671050-e8aee6ee-0742-11e8-8a19-44176c3b0e16.png)

Natim:
It is not a bug it is actually a feature made in #639

PR #637, opened 2018-01-31
User email stored to local storage after enabling sync and removed after disabling sync

himanish-star:
fixes; #629 
GIF of the changes:
ezgif com-video-to-gif

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -103,6 +103,7 @@ ClassicEditor.create(document.querySelector('#editor'), {
 
       chrome.runtime.onMessage.addListener(eventData => {
         let content;
+        let userEmail;
         switch (eventData.action) {
           case 'sync-authenticated':
             setAnimation(true, true, false); // animateSyncIcon, syncingLayout, warning
@@ -111,12 +112,17 @@ ClassicEditor.create(document.querySelector('#editor'), {
             clearTimeout(loginTimeout);
             // set title attr of footer to the currently logged in account
             footerButtons.title = eventData.profile && eventData.profile.email;
+            localStorage.setItem('userEmail', footerButtons.title);
             savingIndicator.textContent = browser.i18n.getMessage('syncProgress');
             browser.runtime.sendMessage({
               action: 'kinto-sync'
             });
             break;
           case 'kinto-loaded':
+            userEmail = localStorage.getItem('userEmail');
+            if(userEmail) {
+              footerButtons.title = userEmail;
+            }
             clearTimeout(loginTimeout);
             content = eventData.data;
             // Switch to Date.now() to show when we pulled notes instead of 'eventData.last_modified'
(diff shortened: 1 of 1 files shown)


src/sidebar/panel.js
```diff
@@ -111,12 +111,16 @@ ClassicEditor.create(document.querySelector('#editor'), {
             clearTimeout(loginTimeout);
             // set title attr of footer to the currently logged in account
             footerButtons.title = eventData.profile && eventData.profile.email;
+            localStorage.setItem('userEmail',footerButtons.title);
```
Natim: nit: add a space after the comma

himanish-star: This is a very stupid question but what is `nit`. And I'll surely make the changes

Natim: nit stand for nit-pick, it is a non really import comment.
Instead that in this case the linter might fail.

src/sidebar/panel.js
```diff
@@ -177,6 +181,9 @@ ClassicEditor.create(document.querySelector('#editor'), {
             syncingInProcess = false;
             break;
           case 'disconnected':
+            if (localStorage.getItem('userEmail')) {
+              localStorage.removeItem('userEmail');
```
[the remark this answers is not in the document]
himanish-star: I'll make the changes in a while.  Currently I don't have my laptop 😅

src/sidebar/panel.js
```diff
@@ -177,6 +183,7 @@ ClassicEditor.create(document.querySelector('#editor'), {
             syncingInProcess = false;
             break;
           case 'disconnected':
+            localStorage.removeItem('userEmail');
```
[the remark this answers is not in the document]
himanish-star: I have made all the changes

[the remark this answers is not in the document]
Natim: It seems to be the case here: https://github.com/mozilla/notes/pull/637/files#diff-1abf0f3e5a69394934ea8f7b5c521805R188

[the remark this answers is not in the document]
himanish-star: Yep, I had seen this but I thought as this eventually leads to the `kinto-loaded case` I thought of making the changes here

src/sidebar/panel.js
```diff
@@ -39,6 +39,7 @@ let ignoreNextLoadEvent = false;
 let ignoreTextSynced = false;
 let lastModified;
 let lastGood = null;
+let userEmail;
```
Natim: I think we should not use a global variable here.

src/sidebar/panel.js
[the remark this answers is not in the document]
Natim: Ok I understood you question, I guess the mistake is to have userEmail as a global variable.

src/sidebar/panel.js
himanish-star: Done

himanish-star: @Natim , I'm so sorry in a hurry I made a forced update and missed out the review commit added by you. I'll add that back

himanish-star:
@Natim, I guess this won't work as `let userEmail` is added somewhere else apart from where it is used. That's the reason why I was making the change. I have the commit ready shall I file a new PR now

Natim:
No I fixed it here: 4b22218

Thanks

vladikoff:
Thank you @himanish-star !!

Issue #574, opened 2018-01-11
Last synced timestamp should be clickable and black


ryanfeeley:
Unlike the last saved timestamp, which is merely informational, the last synced timestamp is both informational and interactive.

Please make sure that it is clickable and black, as designed in the spec.
https://mozilla.invisionapp.com/share/6VBUYHMRB#/235284916_Desktop_Sidebar

Natim:
It is clickable since #558

ryanfeeley:
@Natim And black?

willbmartin:
As my first issue, I would like to contribute to this ticket, and make the text black! Still needed?

ryanfeeley:
@willbmartin Great! I'm doing design on this project and would love the extra help! I thought it was, but then a recent release candidate showed that it was not. I filed this issue more recently, kind of a duplicate I suppose. https://github.com/mozilla/notes/issues/621

Natim:
@willbmartin Yes please go ahead 👍

willbmartin:
Submitted a pull request 👍

PR #639, opened 2018-01-31
sync footer fixed

anweshknayak:
fixes #621

@Natim @vladikoff  please review

I tried changing the property in notes.property but it didn't seem to work so I did it this way [link](https://github.com/mozilla/notes/pull/639/files#diff-1abf0f3e5a69394934ea8f7b5c521805R122)

PR #644, opened 2018-01-31
last sync text color fixed (#574)

willbmartin:
Responding to issue #574 , added styling to make the "Synced at (time)" text black. First pull request, let me know how this looks!

### DIFF
--- a/src/sidebar/editor.js
+++ b/src/sidebar/editor.js
@@ -171,10 +171,10 @@ function setAnimation( animateSyncIcon = true, syncingLayout, warning, syncSucce
     footerButtons.classList.remove('warning');
   }
 
-  if (syncSuccess === true && !footerButtons.classList.contains('syncSuccess')) {
-    footerButtons.classList.add('syncSuccess');
-  } else if (footerButtons.classList.contains('syncSuccess')) {
-    footerButtons.classList.remove('syncSuccess');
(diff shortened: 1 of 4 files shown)


Natim:
Some of this were fixed with #639 can you rebase and remove the .black class from #639 to use your solution instead?

src/sidebar/editor.js
```diff
@@ -170,6 +170,12 @@ function setAnimation( animateSyncIcon = true, syncingLayout, warning ) { // ani
   } else if (warning === false && footerButtons.classList.contains('warning')) {
     footerButtons.classList.remove('warning');
   }
+
+  if (syncSuccess === true && !footerButtons.classList.contains('syncSuccess')) {
+    footerButtons.classList.add('syncSuccess');
+  } else if (footerButtons.classList.contains('syncSuccess')) {
```
Natim: Can we change the if clause to have a else clause here?

src/sidebar/styles.css
```diff
@@ -126,6 +126,10 @@ footer {
   color: black;
 }
 
+#footer-buttons.syncSuccess div:first-child {
```
Natim: I guess we can add the condition close in the previous CSS rule line 125.

willbmartin:
@Natim, on it!

Issue #536, opened 2018-01-03
Add selection context menu to send selection to note


sebastienbarbier:
![contextmenusendtonotes](https://user-images.githubusercontent.com/229453/34514986-47d75dfc-f070-11e7-987c-1015fbd54a7b.png)

vladikoff:
@cedricium if you have time, do you wanna check this one out? See if this works with Sidebar webextensions?

cedricium:
@vladikoff sure thing! 👍

PR #655, opened 2018-02-02
'Send to Notes' context menu

cedricium:
This is for #536 - already had a branch for the 'send-to-notes' context menu and a basic working example (see GIF below). Some work still needs to be done, namely capturing the format of the selected text in order to have that properly displayed in Notes.

**Tasks:**
- [ ] capture and successfully transmit selected text's format
- [x] implement sidebar opening if closed before writing to notepad - e9b5030 
- [x] add GA metrics to this event - 982f2e4 
- [x] clean up code via linting
- [x] resolve merge conflicts - ee5bfa3

**Current Progress:**
send_to_notes


vladikoff:
> namely capturing the format of the selected text

This is interesting. For the intro text we are able to paste HTML into the pad, but I wonder if the context menu is able to grab the HTML formatting, or is it raw text.

vladikoff:
@cedricium what are the chances the context menu can open the side bar ("Send to notes" if sidebar is closed scenario)? Is it capable of doing that? If we can then @ryanfeeley can decide if we want to do that or no.

cedricium:
> This is interesting. For the intro text we are able to paste HTML into the pad, but I wonder if the context menu is able to grab the HTML formatting, or is it raw text.

@vladikoff  this was the hardest part I originally ran into - the object that's returned for the selection context menu only contains a `selectionText` key which is plain-text, not hmtl. Any ideas on how to get the actual html, or would adding the plain-text to the notepad be sufficient?

> what are the chances the context menu can open the side bar ("Send to notes" if sidebar is closed scenario)? Is it capable of doing that? If we can then @ryanfeeley can decide if we want to do that or no.

This shouldn't be too difficult - I think before sending the selected text to `panel.js` we check to see if the sidebar is open and if not, call `browser.sidebarAction.open()`. I'll see what I can do. I think this is a good idea though because if the sidebar is closed, the text that was selected won't be written to Notes due to a communication issue between `background.js` and `panel.js`.

vladikoff:
@cedricium let's see what the simplest thing we can do to ship this feature: 

basic text copy + metrics events + (maybe sidebar auto-opening if it is easy, if too hard then we can skip it)

cedricium:
@vladikoff this should be ready for review. GA metrics have been attached and the sidebar will open to Notes if its closed before adding the selected text. As for the selected text, it will not have its original formatting due to limitations in the API.

src/background.js
```diff
@@ -196,3 +199,20 @@ browser.storage.local.get()
 browser.browserAction.onClicked.addListener(() => {
     browser.sidebarAction.open();
 });
+
+// context menu for 'Send to Notes'
+browser.contextMenus.create({
+  id: 'send-to-notes',
+  title: 'Send to Notes',
```
ryanfeeley: Good label!

src/sidebar/panel.js
```diff
@@ -333,3 +337,17 @@ document.addEventListener('DOMContentLoaded', getLastSyncedTime);
 // Create a connection with the background script to handle open and
 // close events.
 browser.runtime.connect();
+
+function insertSelectedText(editor, text) {
+  const currentNoteContents = editor.getData();
+  let updatedNotesContents = '';
+  if (currentNoteContents === '<p>&nbsp;</p>') {
```
sebastienbarbier: What about if `currentNoteContents` contains the initial content ? My guess is right now it will add `text` at the end, while we probably want to remove it first.

src/background.js
```diff
@@ -159,6 +159,9 @@ browser.runtime.onMessage.addListener(function(eventData) {
     case 'metrics-limit-reached':
       sendMetrics('limit-reached', eventData.context);
       break;
+    case 'metrics-context-menu':
+      sendMetrics('context-menu', eventData.context);
```
vladikoff: Need to add this new event to https://github.com/mozilla/notes/blob/master/docs/metrics.md

src/background.js
sebastienbarbier: Might be nice to integrate i18n to translate `title: 'Send to Notes'` before merging.

src/background.js
```diff
@@ -196,3 +199,20 @@ browser.storage.local.get()
...
+
+// context menu for 'Send to Notes'
+browser.contextMenus.create({
+  id: 'send-to-notes',
+  title: 'Send to Notes',
+  contexts: ['selection']
+});
+
+browser.contextMenus.onClicked.addListener((info) => {
+  browser.sidebarAction.open();
```
vladikoff: @cedricium this doesn't work for me in 58.0.1. How about you?

I get 

> Unchecked lastError value: Error: Could not establish connection. Receiving end does not exist.

src/sidebar/panel.js
vladikoff: could you add a comment regarding why we need to do this here

vladikoff: @ryanfeeley do we want to remove the initial context first? (if it is a fresh pad)

vladikoff:
from meeting: "Can we file a bugzilla issue to allow html formatting for this API?"

src/background.js
cedricium: 🤔 it works for me. I introduced the `contextMenus` permission, maybe a fresh install is needed (`npm run build-ck`)?

cedricium:
@vladikoff were you able to resolve the issue? Also, as for the `insertSelectedText()` function, I've refactored it to just paste the selected text into Notes rather than worrying about empty/newlines.

**Waiting on:**
- [ ] should we remove initial Notes message if present before pasting? cc @ryanfeeley 
- [ ] resolve 'Send to Notes' error @vladikoff saw (could not reproduce)
- [x] 'Send to Notes' should utilize i18n - 6742807

locales/en-US/notes.properties
```diff
@@ -107,3 +107,6 @@ textDirectionTitle=Text Direction
 themeLegend=Theme
 defaultThemeTitle=Default
 darkThemeTitle=Dark
+
+# Context Menu option text
```
flodolo: Suggested comment:

```
# Displayed in the context menu when selecting text on a web page, used to send
# the text to Firefox Notes. Keep "Notes" in English, since it's the name of the
# add-on.
```

vladikoff:
>  resolve 'Send to Notes' error @vladikoff saw (could not reproduce)

Works for me on my other machine, @sebastienbarbier could you do final review?

vladikoff:
@cedricium 

> should we remove initial Notes message if present before pasting? 

I asked @ryanfeeley , he said "No" 👍

docs/metrics.md
```diff
@@ -152,6 +152,12 @@ An event fired when the migration from Quill.JS to CKEditor already happened
 - `ec` - `notes`
 - `ea` - `metrics-migrated-before`
 
+#### `metrics-context-menu`
```
vladikoff: this is gonna be just `context-menu` i think, per `background.js` file

locales/en-US/notes.properties
vladikoff: Need to update this using the suggested comment

src/background.js
```diff
@@ -196,3 +199,20 @@ browser.storage.local.get()
...
+  setTimeout(() => {
+    chrome.runtime.sendMessage({
+      action: 'send-to-notes',
+      text: info.selectionText
+    });
+  }, 700);
```
vladikoff: what is the this `700` magic number? would this cause issues on slower machines? Might be worth commenting this code for the future :)

src/sidebar/panel.js
```diff
@@ -333,3 +337,12 @@ document.addEventListener('DOMContentLoaded', getLastSyncedTime);
 // Create a connection with the background script to handle open and
 // close events.
 browser.runtime.connect();
+
+function insertSelectedText(editor, selectedText) {
+  const currentNotesContent = editor.getData();
+  const updatedNotesContents = currentNotesContent + selectedText;
```
vladikoff: nit: `updatedNotesContents` -> `updatedNotesContent`

vladikoff: because right above you have `currentNotesContent` (singular)

src/background.js
vladikoff: @cedricium can you do:

```js
browser.sidebarAction.open().then(() => {
   chrome.runtime.sendMessage({....})
   // rest of the code 
})
```

cedricium: Doesn't work - the Promise resolves quicker than the actual sidebar opening. Somehow I need to:

1. call `browser.sidebarAction.open()`
2. verify sidebar is open **AND** the ckeditor instance is ready
3. wait for the above step before...
4. sending message to `panel.js` to paste selected text

Let me think about this, I'll see if I can figure it out.

vladikoff: When CKEditor is ready it fires an event to background (i think it already does that). maybe it can fetch back the text from the `selectionText`

cedricium: Do you know where that event is registered in or listened for in either `background.js` or `panel.js`?

vladikoff: https://github.com/mozilla/notes/blob/master/src/sidebar/panel.js#L218

vladikoff: there is a `kinto-load` via loadContent that gets called when editor is ready i think

Natim:
There is a PR to merge master in here: https://github.com/cedricium/notes/pull/2

docs/metrics.md
vladikoff: so in Google analytics it will display the event as `'context-menu'` because panel js sends a ``metrics-context-menu`` and that calls `sendMetrics` `'context-menu'`.

(which later calls `return analytics.sendEvent('notes', event, metrics);` where `event=context-menu`

cedricium:
@vladikoff this should be good, I removed the magic number and instead rely on communication between `Editor.js` and `background.js` to control a `isEditorReady` boolean. That determines whether or not `background.js` can send the selection text. I added comments where relevant to help explain what's going on, see this commit: 13f222c.

flodolo:
@cedricium 
This patch looks great from a localization point of view, but since we switched to the new Fluent syntax yesterday, would you be up for taking it a step further? I can even add a commit on top of your branch if you prefer.

Here's the idea: the new syntax support [terms](https://mozilla-l10n.github.io/localizer-documentation/tools/fluent/basic_syntax.html#basic-syntax), which are a cool feature to use for brand names that should not be translated. So, you can add a couple of terms at the beginning of the file (this is a one-off operation).

```properties
## “Notes” and “Firefox” must be treated as a brand, and kept in English.
## They cannot be:
## - Declined to adapt to grammatical case.
## - Transliterated.
## - Translated.
-brand-notes = Firefox Notes
-brand-notes-short = Notes
```

Those with two `#` are section comments, meaning they will be displayed to all strings in localization tools (Pontoon), until a new section comment appears.

Then, the new string would become

```properties
# Displayed in the context menu when selecting text on a web page, used to send
# the text to Firefox Notes.
sendToNotes = Send to { -brand-notes-short }
```

CC @pike

cedricium:
Hey @flodolo,
Pretty neat feature! Am I missing something however? I updated the `notes.properties` file with the above changes and when viewing the context menu option, the literal text is `Send to { -brand-notes-short }`. The changes I made can be found here: 87acb67 

**Image:**
screen shot 2018-02-15 at 11 20 59 pm

locales/en-US/notes.properties
```diff
@@ -1,4 +1,12 @@
-welcomeTitle3=Welcome to Firefox Notes!
+## “Notes” and “Firefox” must be treated as a brand, and kept in English.
+## They cannot be:
+## - Declined to adapt to grammatical case.
+## - Transliterated.
+## - Translated.
+-brand-notes = Firefox Notes
+-brand-notes-short = Notes
+
+welcomeTitle3=Welcome to { -brand-notes }!
```
flodolo: Please keep this one as it was. To change it, you would need a different ID (e.g. welcomeTitle4), and it would invalidate all existing translations.

The idea is to use the new approach for new strings.

flodolo:
> Pretty neat feature! Am I missing something however? I updated the notes.properties file with the above changes and when viewing the context menu option, the literal text is Send to { -brand-notes-short }. The changes I made can be found here: 87acb67

Uhm, that definitely shouldn't happen. Trying to take a look.

locales/en-US/notes.properties
cedricium: Makes sense, will revert that back.

flodolo:
> Uhm, that definitely shouldn't happen. Trying to take a look.

Sorry, I'm a complete idiot. I'm clearly juggling too many projects these days :-\

This project doesn't use Fluent, uses .properties, which means this feature is not available. Please revert the change. 

Again, sorry for the confusion.

vladikoff:
Awesome! Thanks @cedricium

PR #657, opened 2018-02-02
open/close detection fix

### LINKED ISSUE #656: Always open the notepad when the notes button is pressed
When we press 
image

let's always open the pad.

If users want to hide it let them just press the `X`. 
The current behaviour with the open / closed detection is too messy.

anweshknayak:
fix #656 
@vladikoff please review

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -193,12 +193,5 @@ browser.storage.local.get()
 
 // Handle onClick event for the toolbar button
 browser.browserAction.onClicked.addListener((e) => {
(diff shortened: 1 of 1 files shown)


vladikoff:
@anweshknayak this should remove all code related to `addonIsClosedForWindow` and final version should probably be something like 

```js
browser.browserAction.onClicked.addListener(() => {
  browser.sidebarAction.open();
});
```

anweshknayak:
@vladikoff thanks a lot :smile:

Natim:
Thanks

Issue #189, opened 2017-08-04
Add a view-source / export feature


MasterTuto:
It would be interesting to show the "source code", such as clicking and appearing HTML code, BBCode and also Markdown formats to use on the web.

sebastienbarbier:
Add an export menu to be able to save the content of notes in HTML and/or Markdown. ref #368

PR #674, opened 2018-02-12
WIP: Export Notes content to file

cedricium:
**Overview:**
This is a proof of concept attempt at implementing #189. A menu item has been added in the material design lite menu with a text context of "Export". When clicked, a file titled "notes.md" will be downloaded to the user's computer. The contents of "notes.md" is the contents of the Notes notepad, retrieved using the CKEditor5 `editor.getData()` method. This outputs the HTML of the editor contents - for example, 'Hello World' with no formatting would be `<p>Hello World</p>` in "notes.md".

**Example:**
notes_export

Here is the Notes welcome message downloaded using the "Export" button: https://github.com/cedricium/test/blob/master/notes_html.md

Looking at the raw file, you'll see that it's just HTML: https://raw.githubusercontent.com/cedricium/test/master/notes_html.md

So any editor that allows GitHub-flavored Markdown (GFM) or more generally HTML-to-Markdown support should theoretically be able to view and edit this file.

**What's Next:**
- [x] Menu item "Export" --> "Export as..." with choice between HTML or Markdown
  - should be able to easily parse `editor.getData()`'s HTML output to Markdown using [Turndown](https://github.com/domchristie/turndown) or [Europa](https://github.com/NotNinja/europa) - 54c31ec 
- [ ] design proposal for Export UI from @ryanfeeley 
- [ ] localization of strings "Export as...", "HTML", "Markdown", etc.
- [ ] GA metrics
- [ ] Determine download's `conflictAction` value: "uniquify", "override", or "prompt"
  - maybe append Date/Time string to end of filename to make it unique?

### DIFF
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -291,18 +291,21 @@ disconnectSync.addEventListener('click', disconnectFromSync);
 function exportNotes(editor, fileType) {
   let notesContent = null;
   let exportedFileName = null;
+  let exportFileType = null;
   switch (fileType) {
     case 'html':
       notesContent = editor.getData();
-      exportedFileName = 'notes_html.md';
+      exportedFileName = 'notes.html';
+      exportFileType = 'text/html';
       break;
     case 'markdown':
       notesContent = turndownService.turndown(editor.getData());
-      exportedFileName = 'notes_markdown.md';
+      exportedFileName = 'notes.md';
+      exportFileType = 'text/markdown';
       break;
   }
 
-  const data = new Blob([notesContent], {'type': 'text/markdown'});
+  const data = new Blob([notesContent], {'type': exportFileType});
   const exportFilePath = window.URL.createObjectURL(data);
   const downloading = browser.downloads.download({
     url: exportFilePath,
--- a/src/sidebar/panel.js
+++ b/src/sidebar/panel.js
@@ -12,9 +12,14 @@ const giveFeedbackMenuItem = document.getElementById('give-feedback');
 const savingIndicator = document.getElementById('saving-indicator');
 savingIndicator.textContent = browser.i18n.getMessage('changesSaved');
 
-let FILE_PATH = '';
-const exportButton = document.getElementById('export');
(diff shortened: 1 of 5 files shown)


cedricium:
Two different buttons for choosing which kind of output user wants (proof of concept):
screen shot 2018-02-11 at 11 59 08 pm

**HTML Output:**
https://raw.githubusercontent.com/cedricium/test/master/notes_html.md

**Markdown Output:**
https://raw.githubusercontent.com/cedricium/test/master/notes_markdown.md

src/sidebar/panel.js
```diff
@@ -271,6 +288,28 @@ function disconnectFromSync() {
 
 disconnectSync.addEventListener('click', disconnectFromSync);
 
+function exportNotes(editor, fileType) {
+  let notesContent = null;
+  let exportedFileName = null;
+  switch (fileType) {
+    case 'html':
+      notesContent = editor.getData();
+      exportedFileName = 'notes_html.md';
```
Natim: Let's call it `notes.html`

src/sidebar/panel.js
```diff
@@ -271,6 +288,28 @@ function disconnectFromSync() {
...
+      break;
+    case 'markdown':
+      notesContent = turndownService.turndown(editor.getData());
+      exportedFileName = 'notes_markdown.md';
```
Natim: Let's call it `notes.md`

cedricium:
@ryanfeeley material design lite has an option to disable menu items as well as add a divider under an item - I made a concept of what an updated menu could look like for this Export feature:

notes_mdl_menu

ryanfeeley:
I'm worried that if we bury the Sync icon into the menu, no one will see it.

Can we put the ⫶ to the right of this?
image

I like the divider but we don't need section headers.

How about:

> Export as HTML…
> Export as Markdown…
> 
> Give Feedback… (add ellipsis)

At first I thought that "Sync Your Notes" was to sync manually. I now realize that my text choice is poor. Can we change "Sync Your Notes" to "Enable Sync"?

PR #679, opened 2018-02-13
"Sync with..." tooltip is wrongly displayed when hovering the "Give Feedback" button

himanish-star:
I think this is the only change which I could think of. Please see


vladikoff:
Good fix! Thank you @himanish-star !

PR #693, opened 2018-02-16
Move insertSelectedText to the editor.js lib file.

Natim:


src/sidebar/app/editor.js
```diff
@@ -45,6 +45,16 @@ function customizeEditor(editor) {
   localizeEditorButtons();
 }
 
+function insertSelectedText(editor, selectedText) {
+  const currentNotesContent = editor.getData();
+  const updatedNotesContent = currentNotesContent + `<p>${selectedText.replace(/\n\n/g, '</p><p>')}</p>`;
```
Natim: Also a fix here to replace newlines with a new paragraph.

vladikoff: cc @cedricium just for your info

cedricium: I like that 👍, the content always being on a newline was kinda annoying.

PR #695, opened 2018-02-16
Export as HTML

Natim:
Revamp of #674 

We will go with the **Export as HTML** only for now and wait for a discussion about the Markdown feature altogether before implementing its export.


Natim:
@sebastienbarbier How can I access the editor content from the footer menu? Should I wait for redux?

locales/en-US/notes.properties
```diff
@@ -112,3 +112,5 @@ darkThemeTitle=Dark
 # the text to Firefox Notes. Keep "Notes" in English, since it's the name of the
 # add-on.
 sendToNotes=Send to Notes
+
+exportAsHTML=Export as HTML…
```
vladikoff: @flodolo is this okay?

Issue #699, opened 2018-02-19
"Export as HTML..." tooltip is not displayed when hovering the "Export as HTML..." button


SoftVision-EmilPasca:
[Notes]:
- After logging in to Firefox Notes, "Sync with ..." tooltip is wrongly displayed when hovering the "Export as HTML..." button.

[Affected versions]:
- Firefox 58.0.2 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (3.0.0 installed from https://1228-91073350-gh.circle-artifacts.com/0/tmp/circle-artifacts.MicZiHM/firefox_notes-3.0.0.zip)

[Steps to reproduce]:
1. Open Firefox browser with the profile from prerequisites. 
2. Hover the "Export as HTML..." button from the "Notes" three dots menu.
3. Observe the "Export as HTML..." tooltip.

[Expected result]:
- The "Export as HTML..." tooltip is displayed.

[Actual result]:
- The "Export as HTML..." tooltip is not displayed.

[Additonal Notes]:
- Here is a screen recording with this issue:
![notes_tooltip_exporthtml](https://user-images.githubusercontent.com/20083658/36377837-35f83aa8-1581-11e8-9f28-48f6ae006020.gif)

PR #701, opened 2018-02-19
698 missing sync tooltip

sebastienbarbier:
Moved tooltip in state object so we can easily customize it.
Also STATES is not part of FooterComponent to have direct access to state and read email or last modified date.


Natim:
I guess this is going to remove the tooltip in #699 too. I can still file my PR about adding one there.

Issue #705, opened 2018-02-20
Content can't be erased from "Notes" after reaching the max line limit


SoftVision-EmilPasca:
[Notes]:
- After restarting the browser you can successfully erase content from "Notes".

[Affected versions]:
- Firefox 58.0.2 and up

[Affected Platforms]:
- All Windows
- All Mac
- All Linux

[Prerequisites]:
- Have a Firefox profile with the latest "Firefox Notes" add-on version (3.0.0rc2 installed from https://github.com/mozilla/notes/pull/697#issuecomment-366712158)

[Steps to reproduce]:
1. Open the Firefox browser with the profile from prerequisites. 
2. Select any text and continue pasting it until you reach the max line limit.
3. Erase any text using the "Backspace"/"Delete" keys or "Context-menu" delete option.
4. Observe the "Notes" content.

[Expected result]:
- The text is succesfully erased.

[Actual result]:
- The text cannot be erased.

[Additional Notes]:
- Here is a screen recording with this issue:
![notes_limit_issue](https://user-images.githubusercontent.com/20083658/36418528-c9f8adb2-1637-11e8-83eb-4fd8d51cbefd.gif)

sebastienbarbier:
Aie, bug was introduced during the react refactoring.
We removed all code related to Quill/CKEditor migration, but that included the notification panel which is also used to warn the user about reaching `maximumPadSizeExceeded`.
This need to be re-implemented, I'll do it right away !

PR #706, opened 2018-02-20
Fix broken max line limit in react version

sebastienbarbier:
Fixes https://github.com/mozilla/notes/issues/705

- [x] Can not edit content after reaching max line limit (#705)
- [x] Re-implement notification to warn user about reaching max line limit

### DIFF
--- a/src/sidebar/app/components/Editor.js
+++ b/src/sidebar/app/components/Editor.js
@@ -1,5 +1,7 @@
 import React from 'react';
 
+import CloseIcon from './CloseIcon';
+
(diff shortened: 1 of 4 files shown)


src/sidebar/app/components/Editor.js
```diff
@@ -11,6 +13,7 @@ class Editor extends React.Component {
     this.state = {
       ignoreNextLoadEvent: false,
       ignoreTextSynced: false,
+      isLimitReached: false,
```
Natim: wasLimitReached

PR #730, opened 2018-02-28
Allow Notes to be opened in New Tab

### LINKED ISSUE #470: Provide a button to open notes in a Tab

cedricium:
## Description:

This fixes #470. Adds a new settings option, "Open Notes in..." which allows users to choose either to open Notes in the sidebar or in a new tab. The current implementation will open Notes in the tab when: 

1) the browser action button is pressed, or
2) the "Send to Notes" context menu option is selected.

Before opening a new Notes tab, all of the tabs open will be queried to see if a Notes tab exists - if so, then there will be no action, making sure there are no duplicate Notes tabs.

The **edge cases that will still open / show Notes in the sidebar** despite the "Open Notes in new tab" option being selected are **when the keyboard shortcut is used**, or when **Notes is manually selected from the dropdown menu in the sidebar.**

**cc:** @vladikoff @ryanfeeley let me know if this is what you had in mind.

### Settings Page:
<div align="center">
  
</div>

### "Open Notes in New Tab" Demo:
<div align="center">
   
</div>

### DIFF
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -108,6 +108,10 @@ themeLegend=Theme
 defaultThemeTitle=Default
 darkThemeTitle=Dark
 
+locationLegend=Open Notes in...
+defaultLocationLabel=Sidebar
+tabLocationLabel=New Tab
+
 # Displayed in the context menu when selecting text on a web page, used to send
 # the text to Firefox Notes. Keep "Notes" in English, since it's the name of the
 # add-on.
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -108,7 +108,7 @@ themeLegend=Theme
 defaultThemeTitle=Default
 darkThemeTitle=Dark
 
-locationLegend=Open Notes in...
+locationLegend=Open Notes in
 defaultLocationLabel=Sidebar
 tabLocationLabel=New Tab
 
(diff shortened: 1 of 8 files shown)


locales/en-US/notes.properties
```diff
@@ -108,6 +108,10 @@ themeLegend=Theme
 defaultThemeTitle=Default
 darkThemeTitle=Dark
 
+locationLegend=Open Notes in...
```
flodolo: As a general note, always use `…` instead of `...`

Having said that, `…` seems wrong for the context. It's not a button label requiring more info from the user or opening a dialog, it's a label for the two radio options.

cedricium: Thanks @flodolo, I've gone ahead and removed the ellipsis entirely.

sebastienbarbier:
Wow that's awesome 👍 !
I really like it and cannot wait to see feedbacks from UX.

vladikoff:
@cedricium not sure if this is the only styling issue 
image

The sync arrow is lower, not even sure why 😯

cedricium:
Good find @vladikoff! I didn't really look at the editor too much when it was in the tab, I'll look for other styling issues as well.

cedricium:
@vladikoff styling issues have been resolved 👍

PR #731, opened 2018-02-28
Fix browser-specific `start` scripts

cedricium:
## Problem:

With the way the current `npm run start-nightly` script is written, Firefox Nightly won't actually run - rather, regular Firefox opens.

```sh
# `npm run start-nightly`
> notes@3.1.0-dev start-nightly /Users/cedricamaya/Documents/GitHub/notes
> npm start -- --firefox=nightly


> notes@3.1.0-dev start /Users/cedricamaya/Documents/GitHub/notes
> npm run webpack && web-ext run -s build & npm run webpack:watch "--firefox=nightly"
```

Notice the `"--firefox=nightly"` isn't part of the `web-ext run` command, resulting in regular FF to start.

## Solution:

Instead of using the `npm start` command in browser-specific scripts (`npm run start-nightly` and `npm run start-deved`), explicitly call `npm run webpack...` similar to `npm run start`.


cedricium:
@vladikoff @Natim @sebastienbarbier have you ran into this issue? Maybe it's just me...

vladikoff:
This makes sense, @sebastienbarbier did you have any feedback to avoid repetition in the scripts or should we just leave it at that?

sebastienbarbier:
I've been thinking about this but couldn't find anything I was happy with, so I guess we can leave it like this for now. 👍

vladikoff:
👍  Thanks @cedricium !

Issue #735, opened 2018-02-28
Add current date and time to Export as HTML


vladikoff:
### [Description]:

Add current date and time to Export as HTML

### [Steps to reproduce]:

Export notes as HTML via the menu

### [Expected result]:

`notes-2018-02-28-10_38am.html`

@ryanfeeley can you suggest formatting on the file for MAXIMUM UXing?

### [Actual result]:
![image](https://user-images.githubusercontent.com/128755/36796416-85133fb8-1c73-11e8-9716-4b029b3e1a37.png)

ryanfeeley:
@vladikoff Looking at my screenshots, this 24:00 clock format seems to work well. Sorts nicely.

``Firefox Notes 2017-12-24 21.23.31.html``

vladikoff:
@cedricium this fix we will land for sure ^ if you have time

cedricium:
@vladikoff I can have this in at the end of the day (Pacific time) 👍

@ryanfeeley you want the spaces included in the filename?

ryanfeeley:
@cedricium My Windows knowledge is extremely poor. I know that Dropbox includes spaces and periods on the Mac client. Not sure how big a sin this is on Windows and/or Linux.

ryanfeeley:
@cedricium OK, I spoke with the higher powers. We should follow was Firefox does when you **Save Page As…** and just use the title of the note.

e.g ``Mattress Shopping.html`` no time stamp

If the note is blank use: ``blank.html`` or better, disable the export note option in that state.

vladikoff:
> We should follow was Firefox does when you Save Page As… and just use the title of the note.

@ryanfeeley we don't have multi note with this yet though, do you want to get the first Heading or first sentence of the notepad?

vladikoff:
Almost done in https://github.com/mozilla/notes/pull/738

Issue #733, opened 2018-02-28
The export file is not displayed in the downloads panel if it is downloaded in a Private Window


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 58.0.2 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (3.1.0dev custom built from 28.02.2018) installed.
- Have the sidebar opened.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites and open a New Private Window.
2. Close the Normal Window (non-private).
3. Click the "Three dots" button from the bottom-right part of the sidebar.
4. Choose the "Export as HTML" option.
5. Click the "Download Arrow" button and observe the displayed downloads.

**[Expected result]:**
- The export file is displayed in the downloads panel.

**[Actual result]:**
- No downloads are displayed for this session.

**[Notes]:
- Although the "notes.html" file is not displayed in the "about: downloads" page, the file it was downloaded and can be find in selected folder for downloads.
- If you also have a normal window opened then the export file will be displayed in the normal window's downloads panel, instead of the private window's one.
- The issue is reproducible with the TxP version of "Firefox Notes".
- Attached a screen recording of the issue: 
![rec of the issue](https://user-images.githubusercontent.com/31243259/36795182-050cf2ea-1cab-11e8-87d7-1ac469c397eb.gif)

vladikoff:
@ryanfeeley I guess Firefox Desktop hides the downloads button in the private window. What do you want to do here?

<img src=https://user-images.githubusercontent.com/128755/36795374-e20c3a42-1c70-11e8-8e8c-bdb4c9e980ca.png width=200 /> 

The photon animation actually happens in the non-Private window.


@Softvision-MariusComan this doesn't happen if this is set

<img src=https://user-images.githubusercontent.com/128755/36795478-22da7dc2-1c71-11e8-934d-c8f90833490a.png width=300 />


<img src=https://user-images.githubusercontent.com/128755/36795407-f93841de-1c70-11e8-9507-366e1be1197b.png width=300 />


so at least this is nice

cedricium:
@vladikoff if we add `saveAs: true` to the following code:

```js
// Footer.js, L165-168
browser.downloads.download({
  url: exportFilePath,
  filename: exportedFileName
});
```
[...]

PR #738, opened 2018-03-01
Use Notes content to name "Export as" filenames

cedricium:
Fixes https://github.com/mozilla/notes/issues/733
Fixes https://github.com/mozilla/notes/issues/735

Eventual fix for #735.

@vladikoff mind going over what I have so far? I feel as if I'm over-complicating things, but at the same time I feel like there's a lot needed to get the correct filename.

Per @ryanfeeley:

> OK, I spoke with the higher powers. We should follow was Firefox does when you Save Page As… and just use the title of the note.
> e.g Mattress Shopping.html no time stamp
> If the note is blank use: blank.html or better, disable the export note option in that state.

and 

> Yes, the first line that has text, it should use the text from that line to a maximum of 255 characters and just end it, no ellipsis.

My thought process was then:

1. **Notes not empty:** Notes has text in the notepad. Said text could possibly be halfway down the page meaning there are multiple empty `<p>` elements before the text:

```html
<p></p>
<p></p>
<p></p>
<p>Some text</p>

<!-- filename = "Some text.html" -->
```
This means we need to find the first element that contains text in order to set it as the filename. This first commit does just this. Still need to only get the first sentence or ensure the text is no longer than 255 characters.

2. **Notes empty:** If Notes is empty, then we could just set the title of the filename to "blank.html" (currently implemented). If we did want to go the "disable the context menu option if Notes empty", I might advise against it. It would be taxing to constantly check if Notes is empty and Google allows for downloading documents without a title.

### DIFF
--- a/src/sidebar/app/components/Footer.js
+++ b/src/sidebar/app/components/Footer.js
@@ -162,7 +162,7 @@ class Footer extends React.Component {
       const parentElement = document.createElement('div');
       parentElement.innerHTML = notesContent; // eslint-disable-line no-unsanitized/property
 
-      let exportFileName =  'blank.html';
+      let exportFileName = 'blank.html';
       // get the first child element with text
       const nonEmptyChildElement = getFirstNonEmptyElement(parentElement);
 
@@ -221,7 +221,7 @@ class Footer extends React.Component {
         const that = this;
         this.loginTimeout = setTimeout(() => {
           that.setState({
-            state:  this.STATES.PLEASELOGIN
+            state: this.STATES.PLEASELOGIN
           });
         }, 5000);
 
--- a/src/sidebar/app/components/Footer.js
+++ b/src/sidebar/app/components/Footer.js
@@ -3,7 +3,7 @@ import classNames from 'classnames';
 
 import SyncIcon from './SyncIcon';
 
-import { formatFooterTime } from '../utils/utils';
+import { formatFooterTime, getFirstNonEmptyElement } from '../utils/utils';
 import { SURVEY_PATH } from '../utils/constants';
 import INITIAL_CONTENT from '../data/initialContent';
 
@@ -156,10 +156,24 @@ class Footer extends React.Component {
     };
 
     this.exportAsHTML = () => {
+      // get Notes content
       const notesContent = this.state.content;
-      const exportedFileName = 'notes.html';
-      const exportFileType = 'text/html';
+      // assign contents to container element for later parsing
+      let parentElement = document.createElement('div');
+      parentElement.innerHTML = notesContent;
 
+      let exportedFileName =  '';
+      // get the first child element with text
+      let nonEmptyChildElement = getFirstNonEmptyElement(parentElement);
+
+      // if non-empty child element exists, set the filename to the element's `textContent`
+      if (nonEmptyChildElement) {
+        exportedFileName = `${nonEmptyChildElement.textContent.trim()}.html`;
+      } else { // if child elements are empty (aka empty notepad), set filename to 'blank.html'
+        exportedFileName = 'blank.html';
+      }
+
+      const exportFileType = 'text/html';
       const data = new Blob([`<!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><title>Notes</title></head><body>${notesContent}</body></html>`], {'type': exportFileType});
       const exportFilePath = window.URL.createObjectURL(data);
       browser.downloads.download({
--- a/src/sidebar/app/components/Footer.js
+++ b/src/sidebar/app/components/Footer.js
@@ -3,7 +3,7 @@ import classNames from 'classnames';
 
 import SyncIcon from './SyncIcon';
 
-import { formatFooterTime, getFirstNonEmptyElement } from '../utils/utils';
+import { formatFooterTime, getFirstNonEmptyElement, formatFilename } from '../utils/utils';
 import { SURVEY_PATH } from '../utils/constants';
 import INITIAL_CONTENT from '../data/initialContent';
 
@@ -159,18 +159,16 @@ class Footer extends React.Component {
       // get Notes content
       const notesContent = this.state.content;
       // assign contents to container element for later parsing
-      let parentElement = document.createElement('div');
-      parentElement.innerHTML = notesContent;
+      const parentElement = document.createElement('div');
+      parentElement.innerHTML = notesContent; // eslint-disable-line no-unsanitized/property
 
-      let exportedFileName =  '';
+      let exportFileName =  'blank.html';
       // get the first child element with text
-      let nonEmptyChildElement = getFirstNonEmptyElement(parentElement);
+      const nonEmptyChildElement = getFirstNonEmptyElement(parentElement);
 
       // if non-empty child element exists, set the filename to the element's `textContent`
       if (nonEmptyChildElement) {
-        exportedFileName = `${nonEmptyChildElement.textContent.trim()}.html`;
-      } else { // if child elements are empty (aka empty notepad), set filename to 'blank.html'
-        exportedFileName = 'blank.html';
+        exportFileName = formatFilename(nonEmptyChildElement.textContent);
       }
 
       const exportFileType = 'text/html';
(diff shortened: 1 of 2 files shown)


vladikoff:
@cedricium Looking at this now! 

(In the future could you please use branches in this repo, under `mozilla/notes`, this way we can all contribute to the same branch. You should have access to create branch on this repo)

vladikoff:
```
/home/travis/build/mozilla/notes/src/fxa-utils.js
  61:18  warning  Expected to return a value at the end of arrow function  consistent-return
/home/travis/build/mozilla/notes/src/sidebar/app/components/Footer.js
  162:11  error  'parentElement' is never reassigned. Use 'const' instead         prefer-const
  163:7   error  Unsafe assignment to innerHTML                                   no-unsanitized/property
  167:11  error  'nonEmptyChildElement' is never reassigned. Use 'const' instead  prefer-const
/home/travis/build/mozilla/notes/src/sidebar/app/utils/utils.js
  15:3   error  Trailing spaces not allowed                                                                                 no-trailing-spaces
  16:45  error  Trailing spaces not allowed                                                                                 no-trailing-spaces
  23:15  error  'index' is assigned a value but never used. Allowed unused vars must match /^Cc|Ci|Cu|Cr|EXPORTED_SYMBOLS/  no-unused-vars
/home/travis/build/mozilla/notes/src/sync.js
  173:7   warning  Arrow function expected no return value                  consistent-return
  192:22  warning  Expected to return a value at the end of arrow function  consistent-return
✖ 9 problems (6 errors, 3 warnings)
  4 errors, 0 warnings potentially fixable with the `--fix` option.

```

src/sidebar/app/components/Footer.js
```diff
@@ -156,10 +156,24 @@ class Footer extends React.Component {
...
     this.exportAsHTML = () => {
+      // get Notes content
       const notesContent = this.state.content;
-      const exportedFileName = 'notes.html';
-      const exportFileType = 'text/html';
+      // assign contents to container element for later parsing
+      let parentElement = document.createElement('div');
+      parentElement.innerHTML = notesContent;
+
+      let exportedFileName =  '';
```
vladikoff: set this to `'blank.html'` right away, to avoid the `else` below on line 173

src/sidebar/app/utils/utils.js
```diff
@@ -11,4 +11,20 @@ function formatFooterTime(date) {
...
+/**
+ * 
+ * @param {HTMLElement Object} parentElement 
+ * @returns {HTMLElement Object}
+ */
+function getFirstNonEmptyElement(parentElement) {
+  // create an Array from parentElement's `children` (HTMLCollection)
+  const parentElementChildrenArray = Array.prototype.slice.call(parentElement.children);
+  // search for first child element that is not empty and return it
+  for (const [index, childElement] of parentElementChildrenArray.entries()) {
```
vladikoff: limit this to like 15 - 25 parentElement children, so avoid issues with big notepads, if we cannot extract title from that then we can just use blank.html

src/sidebar/app/utils/utils.js
```diff
@@ -11,4 +11,20 @@ function formatFooterTime(date) {
...
 
-export { formatFooterTime };
+/**
+ * 
+ * @param {HTMLElement Object} parentElement 
+ * @returns {HTMLElement Object}
+ */
+function getFirstNonEmptyElement(parentElement) {
+  // create an Array from parentElement's `children` (HTMLCollection)
+  const parentElementChildrenArray = Array.prototype.slice.call(parentElement.children);
```
vladikoff: might do the limiting here (per the other comment)

vladikoff:
Fixes https://github.com/mozilla/notes/issues/735

cedricium:
@vladikoff will start using the `mozilla/notes` repo 👍 

If you want to look over this one last time, it should be good to go. I made the changes you requested (automatically set filename to `blank.html` and limit the number of child elements we traverse, plus lint issues).

One thing I did add was a `formatFilename()` function formats the filename by removing illegal filename characters and shortens the filename if need be. Also add the `saveAs: true` to always open the file chooser in order to fix #733.

src/sidebar/app/components/Footer.js
```diff
@@ -156,15 +156,28 @@ class Footer extends React.Component {
...
+      }
+
+      const exportFileType = 'text/html';
       const data = new Blob([`<!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><title>Notes</title></head><body>${notesContent}</body></html>`], {'type': exportFileType});
       const exportFilePath = window.URL.createObjectURL(data);
       browser.downloads.download({
         url: exportFilePath,
-        filename: exportedFileName
+        filename: exportFileName,
+        saveAs: true // always open file chooser, fixes #733
```
vladikoff: 👍

src/sidebar/app/utils/utils.js
```diff
@@ -11,4 +11,45 @@ function formatFooterTime(date) {
...
+ * @param {string} filename
+ * @returns {string}
+ */
+function formatFilename(filename) {
+  let formattedFilename = filename;
+  // remove surrounding whitespace
+  formattedFilename = formattedFilename.trim();
+  // remove illegal filename characters
+  formattedFilename = formattedFilename.replace(/[~#%{}[\]:\\<>/!@&?"*.+|\n\r\t]/g, '');
+  if (formattedFilename.length > 250) { // 255 bytes (filesystem max) - 4 for ".html" extension
```
vladikoff: `.html` is actually 5 bytes. forgot the`.` :)

src/sidebar/app/components/Footer.js
```diff
@@ -156,15 +156,28 @@ class Footer extends React.Component {
...
     this.exportAsHTML = () => {
+      // get Notes content
       const notesContent = this.state.content;
-      const exportedFileName = 'notes.html';
-      const exportFileType = 'text/html';
+      // assign contents to container element for later parsing
+      const parentElement = document.createElement('div');
+      parentElement.innerHTML = notesContent; // eslint-disable-line no-unsanitized/property
+
+      let exportFileName =  'blank.html';
```
vladikoff: there is a double space here `  'blank.html'`

src/sidebar/app/utils/utils.js
```diff
@@ -11,4 +11,45 @@ function formatFooterTime(date) {
...
-export { formatFooterTime };
+/**
+ *
+ * @param {HTMLElement Object} parentElement
+ * @returns {HTMLElement Object}
+ */
+function getFirstNonEmptyElement(parentElement) {
+  // create an Array from parentElement's `children` (HTMLCollection)
+  let parentElementChildrenArray = [];
+  if (parentElement.children.length > 20) {
```
vladikoff: @cedricium do you need this check, can you just `.slice(0,20)` no matter what the size is.

src/sidebar/app/utils/utils.js
```diff
@@ -11,4 +11,45 @@ function formatFooterTime(date) {
...
+    parentElementChildrenArray = Array.prototype.slice.call(parentElement.children[0, 20]);
+  } else {
+
```
vladikoff: no need for this empty new line, but I have a feeling you won't need this `else` anyway

src/sidebar/app/utils/utils.js
```diff
@@ -11,4 +11,45 @@ function formatFooterTime(date) {
...
+    parentElementChildrenArray = Array.prototype.slice.call(parentElement.children);
+  }
+  // search for first child element that is not empty and return it
+  for (const childElement of parentElementChildrenArray.entries()) {
```
vladikoff: this can be cleaner, could probably use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find  ?

src/sidebar/app/utils/utils.js
vladikoff: also do you need to check for > 250, do a substring on MAX_CHARS + 5 bytes of extension

vladikoff: if name is smaller substring will be "ignored" sort of, if it is longer then substring will get it.

cedricium: Oops! Code is right, will need to change the comment lol

src/sidebar/app/utils/utils.js
cedricium: Check is necessary, otherwise if there are less than 20 elements then `slice` will automatically create an empty array.

src/sidebar/app/utils/utils.js
cedricium: Jeez, nice catch! There was a `console.log` statement there and somehow I didn't get rid of the newline :P

PR #777, opened 2018-03-16
Disable "Export" button while Notes syncs

### LINKED ISSUE #767: Disable the "Export as HTML" option while the "Saving changes..." message is displayed
**[Notes]:**
- This issue is much more easily encountered when you are logged in and the "Syncing changes..." message is displayed.

**[Affected versions]:**
- Firefox 58.0.2 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (3.1.0rc2) installed.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Paste or write something in the sidebar.
3. Quickly click the "Notes" menu button.
4. Observe the "Export as HTML" option while the "Saving changes..." message is still displayed.

**[Expected result]:**
- The "Export as HTML" option is inactive.

**[Actual result]:**
- The "Export as HTML" option is active and clickable.

**[Additional Notes]:
- If the "Export as HTML" option is clicked while the "Saving changes..." message is displayed then the downloaded HTML file will not contain the latest changes.
- The issue is reproducible with the TxP version of "Firefox Notes".
rec of the issue

cedricium:
Fixes #767

**Demo:**
notes_disable_export

### DIFF
--- a/src/sidebar/app/components/Footer.js
+++ b/src/sidebar/app/components/Footer.js
@@ -105,6 +105,7 @@ class Footer extends React.Component {
           // Disable sync-action
           break;
         case 'text-editing':
+          document.getElementById('export-button').setAttribute('disabled', true);
           this.setState({
             state: this.state.isAuthenticated ? this.STATES.SYNCING : this.STATES.SAVING
           });
@@ -116,6 +117,7 @@ class Footer extends React.Component {
             content: eventData.content || INITIAL_CONTENT
           });
           this.getLastSyncedTime();
+          document.getElementById('export-button').removeAttribute('disabled');
           break;
         case 'text-saved':
           if (!this.state.state.ignoreChange && !this.state.isAuthenticated) {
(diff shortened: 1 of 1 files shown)


vladikoff:
@sebastienbarbier can you suggest a better way for this fix?

vladikoff:
We might be able to solve this with React, but that would change the state a bit

cedricium:
@vladikoff I agree, we should probably use React. And I doubt it's as simple as disabling / enabling the context menu button with the sync, there are probably other edge cases we need to look out for (that React could solve).

sebastienbarbier:
Redux solve this problem by having a single centralised shared state.
Looking at master right now, could not find an easy fix so might need to spend some time on this.
I guess we could go for disabled button until we release 4.0.

vladikoff:
Thanks @cedricium !!

Issue #822, opened 2018-03-28
Refactor background.js to use webworker


sebastienbarbier:
Webworker have a different global scope and unfortunately our current background code heavily depend on those variable.
We might benefits regarding performance and maintainability to get rid of those.

This issue is created after working on #783

PR #823, opened 2018-03-28
Fix copy/paste delay in editor and build backgound.js using webpack

### LINKED ISSUE #783: Investigate performance syncing long text
### [Description]:

Might be able to use a WebWorker or some other cool web tech (tm)

Extracted from: https://github.com/mozilla/notes/pull/743

sebastienbarbier:
This PR fix copy paste issue on ckeditor. It now use webpack to build background.js.

It try to use webworker, but current background code is **heavily** dependant on global variables and webworker have a different global scope. Meaning it works, but has nothing to run without an intensive time consuming refactoring. Might not be worth based on current timeline (releasing v4).

Using webpack to build background will force us to use less global variables and hopefully help on a futur refactoring. An issue has been created: #822

Fix #783

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -30,7 +30,6 @@ module.exports = function(config) {
     files: [
       "src/vendor/*.js",
       "test/dist/unit-bundle.js",
-      "src/*.js",
     ],
     // coverage preprocessor doesn't cope with ES2017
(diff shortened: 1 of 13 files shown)


vladikoff:
From mtg: let's make a new PR just for the `copy / pasta` fixes

PR #830, opened 2018-03-28
Fix copy paste in ckeditor

sebastienbarbier:


sebastienbarbier:
@vladikoff We should merge that one 👍.

vladikoff:
found a multi window issue in this one

PR #836, opened 2018-03-28
Dropped font weight from 600 to 500

ryanfeeley:
Dropped font weight from 600 to 500 to make it semibold


vladikoff:
Thanks @ryanfeeley !

PR #842, opened 2018-03-29
Add CD10 to all metrics

### LINKED ISSUE #771: Add CD 10: is user syncing
this should be a boolean...we can use this to sort whether syncers retain better.

sebastienbarbier:
Add cd10 to every metrics request. Possible values are: **'error', 'isSyncing', 'synced', 'openLogin', 'verifyAccount', 'reconnectSync', and 'signIn'**.
Names are the one used by front-end to display proper footer UI.

We are not able for now to detect if a user synced on more than one device. This should be an improvement while working on the v5.mobile app.

Fix #771

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -27,8 +27,8 @@ const analytics = new TestPilotGA({
   av: browser.runtime.getManifest().version
 });
 
-// This object is updated onMessage 'redux', sended by sidebar store.js on evry change.
-// We need to keep it to generate cd10
+// This object is updated onMessage 'redux', sended by sidebar store.js on every change.
+// We need to keep it to generate `cd10`
 let reduxSync = {};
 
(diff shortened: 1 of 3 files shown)


src/background.js
```diff
@@ -54,10 +58,31 @@ function sendMetrics(event, context = {}) {
...
+      } else {
+        metrics.cd10 = 'signIn';
+      }
+    }
+
     return analytics.sendEvent('notes', event, metrics);
   };
   clearTimeout(timeouts[event]);
-  timeouts[event] = setTimeout(later, 20000);
+  timeouts[event] = setTimeout(later, 4000);
```
sebastienbarbier: Oups, I did change that value for testing purpose but should stay at 20000. Will fix in a second.

Issue #832, opened 2018-03-28
The "Send to Notes" context menu option no longer works on the latest Beta and Nightly


Softvision-MariusComan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1449614

**[Affected versions]:**
- Firefox Beta (60.0b3) and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (4.0.0 Beta 3) installed.
- The "Firefox Notes" sidebar is opened in the list view mode.
- Have any website opened.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Select any text from the opened website and right click on it.
3. Click the "Send to Notes" option and observe the behavior.

**[Expected result]:**
- The selected text is displayed in the sidebar in a new note.

**[Actual result]:**
- The sidebar flickers and the selected text is not added as a new note.

**[Regression window]:**
Last good revision: 7433031c9274b8299c3e2d50613de74352684bc1
First bad revision: 95a6bd5b90452da6eee11eb8ee53ef80c91565c6
INFO: Pushlog:
https://hg.mozilla.org/integration/mozilla-inbound/pushloghtml?fromchange=7433031c9274b8299c3e2d50613de74352684bc1&tochange=95a6bd5b90452da6eee11eb8ee53ef80c91565c6

**[Additional Notes]:**
- If a note is opened when the "Send to Notes" context menu option is clicked, the sidebar goes back in the list view mode. In this case the text should have been added to the opened note.
- If the sidebar is closed when the "Send to Notes" context menu option is selected, the sidebar is opened in the list view mode. In this case the selected text should have been added as a new note.
- The issue is not reproducible with the TxP version of "Firefox Notes".
- Attached a screen recording of the issue:
![rec of the issue](https://user-images.githubusercontent.com/31243259/38041668-0c158bda-32bb-11e8-9f84-55d0cbb660eb.gif)

vladikoff:
@Softvision-MariusComan I'm super confused because this was supposed to be fixed here: https://bugzilla.mozilla.org/show_bug.cgi?id=1445080

Ref: https://github.com/mozilla/notes/issues/740

mixedpuppy:
A change in notes 4 is that:

- it loads /sidebar/index.html, 
- changes the url to /sidebar/index.html#/
- when you use the context menu it loads /sidebar/index.html.  

[...]

PR #857, opened 2018-03-31
Fix send to notes for react router

vladikoff:
Fixes https://github.com/mozilla/notes/issues/832


sebastienbarbier:
Solved it. Was broken for every version, but basically send-to-note was performed locally and never propageted to kintojs. @vladikoff can you test and let me know if that's good for you 👍

PR #880, opened 2018-04-04
Fix crash on migration while synced

### LINKED ISSUE #879: Upgrading from Notes v3 to v4 while synced crashes the notepad
### [Description]:

upgrade

sebastienbarbier:
PR fix this issue, but we lose authentication after migration since email was not stored locally in v3. Hope that is ok.

Fix #879

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -86,6 +86,18 @@ function sendMetrics(event, context = {}, sync = reduxSync) {
   timeouts[event] = setTimeout(later, 20000);
 }
 
+function fetchProfile(credentials) {
+  fxaFetchProfile(FXA_PROFILE_SERVER, credentials.access_token).then((profile) => {
+    browser.storage.local.set({credentials}).then(() => {
+      chrome.runtime.sendMessage({
+        action: 'sync-authenticated',
+        credentials,
+        profile
+      });
+    });
+  });
+}
+
 function authenticate() {
   const fxaKeysUtil = new fxaCryptoRelier.OAuthUtils();
     chrome.runtime.sendMessage({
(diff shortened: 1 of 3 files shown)


sebastienbarbier:
Should be good 👍, if we have credential but no email we retrieve user profile and authenticate.
UI is optimistic so we load footer as synced and tooltip display during few seconds 'Synced with ...'.

src/sidebar/app/app.js
```diff
@@ -36,9 +36,18 @@ browser.storage.local.get().then(result => {
...
   const state = JSON.parse(result.redux || '{}');
+
   // We use stored state to propagate actions and avoid keeping
-  if (result.hasOwnProperty('credentials') && state.sync && state.sync.email) {
-    store.dispatch(authenticate(state.sync.email));
+  if (result.hasOwnProperty('credentials')) {
+    if (state.sync && state.sync.email) {
+      store.dispatch(authenticate(state.sync.email));
+    } else {
+      store.dispatch(authenticate('...'));
```
vladikoff: `store.dispatch(authenticate('...')` the `...` looks strange, what does that do?

sebastienbarbier: UI is optimistic so we load footer as synced and tooltip display during few seconds 'Synced with ...'.

src/background.js
```diff
@@ -86,6 +86,18 @@ function sendMetrics(event, context = {}, sync = reduxSync) {
   timeouts[event] = setTimeout(later, 20000);
 }
 
+function fetchProfile(credentials) {
+  fxaFetchProfile(FXA_PROFILE_SERVER, credentials.access_token).then((profile) => {
```
vladikoff: this request may fail , for example due to an expired token. 

If this request fails we should just "log out" and user can resignin in that case

src/sidebar/app/app.js
vladikoff: ah cool

src/background.js
vladikoff: basically do we need to catch the error case here or it doesn't matter?

sebastienbarbier: I believe logout will occur because kinto-sync will fail using expired credential but just to be sure I pushed a catch sending `reconnect` message.

sebastienbarbier:
Note for future us, we should consider re factoring background.js and sync.js: many functions and start to be too long 😉 ... (ping #822 )

src/background.js
```diff
@@ -227,7 +227,11 @@ browser.runtime.onMessage.addListener(function(eventData) {
       break;
     case 'fetch-email':
       credentials.get().then(received => {
-        fetchProfile(received);
+        fetchProfile(received).catch(e => {
+          chrome.runtime.sendMessage({
```
vladikoff: yeeee 👍

PR #881, opened 2018-04-04
Fix unexpected welcome note 

### LINKED ISSUE #873: Upgrading from Notes 3 to Notes 4 creates a "welcome note"
### [STR]:

* Install stable notes from https://testpilot.firefox.com/experiments/notes/
* Type `wow notes` into the single note window

* Side-load Notes v4 using `about:debugging`

### [Actual]

* Add-on shows welcome note 
* Going back the notes list I see 2 notes

### [Expected]

* Only my note should appear 
(Should not see Welcome as an existing user) 

image

sebastienbarbier:
We now create initial note only if **first time using redux** AND **kinto is loaded** AND **it has no note**.

WIP, need more testing.

Fix #873

### DIFF
--- a/src/sidebar/app/app.js
+++ b/src/sidebar/app/app.js
@@ -41,7 +41,10 @@ browser.storage.local.get().then(result => {
     store.dispatch(authenticate(state.sync.email));
   }
 
-  store.dispatch(kintoLoad(state.notes ? state.notes : []));
+  if (state.notes) {
+    store.dispatch(kintoLoad(state.notes));
+  }
+
 
   // Render root DOM element
   ReactDOM.render((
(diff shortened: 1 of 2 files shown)


sebastienbarbier:
All right this PR works as expected but code structure is not great...  Would refactor using a loading panel, is that an option @ryanfeeley ? I know @vladikoff did some similar work in our native mobile prototype, which make sense.

vladikoff:
Worked in my testing 👍

ryanfeeley:
@sebastienbarbier I'm not sure I understand. Actually, I'm sure I don't understand.

sebastienbarbier:
@ryanfeeley haha I'm also confused by my question, was probably late when I wrote my question, sorry for that. I was thinking about a loading animation between opening sidebar and displaying listPanel, but let's forget about it for now it should be fine without it 😅.

PR #885, opened 2018-04-04
Copy tweaks mostly title case

ryanfeeley:
A few mismatches from the mockup that I decided to fix or break things on GitHub

### DIFF
--- a/locales/en-US/notes.properties
+++ b/locales/en-US/notes.properties
@@ -53,8 +53,8 @@ removeAccount=Remove Account
 
 openingLogin=Opening login…
 openingLoginWindow=Opening login window…
-pleaseLogin=Verify your account…
-forgetEmail=Forget this Email
+pleaseLogin=Verify Your Account
+forgetEmail=Forget This Email
 disableSync=Disable Sync
 reconnectSync=Reconnect to Sync
 syncError=Sync Error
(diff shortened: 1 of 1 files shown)


vladikoff:
@flodolo r?

locales/en-US/notes.properties
```diff
@@ -95,7 +95,7 @@ title1=Heading 1
 title2=Heading 2
 title3=Heading 3
 paragraph=Body
-bodyText=Body text
+bodyText=Body Text
```
vladikoff: I guess we cannot really change any of this text like. every change needs to introduce a new variable. However these are minimal changes to English that won't matter that much for other locales? @flodolo let us know

sebastienbarbier: What about using CSS like `text-transform: capitalize` on bodyText ? Would avoid a review from i18n team and force styling for every language.

flodolo: Let's keep the existing string IDs, each locale should have their own rules for capitalization.

Using CSS text-transform is a no-go for web projects, where we don't control the rendering engine. 

If the only engine is Gecko, we know it does the right thing with `uppercase`, as long as you pass the correct language code. For the record, that causes bugs in Firefox UI.

In general, it's better to use capitalization in the string and let each locale decide what to use.

`capitalize` is really poor, since it's not proper letter case even in English.

Issue #858, opened 2018-04-02
The Notes List is not correctly synchronized across Firefox profiles when a note is deleted


SoftVision-EmilPasca:
**[Affected versions]:**
- Firefox 59.0.2 and above

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have two Firefox profiles with the latest Notes version (4.0.0 Beta 5) enabled.
- You must be logged with the same "Firefox Notes" account on both profiles.
- Make sure you have the same notes in the "Note List" View on both profiles.

**[Steps to reproduce]:**
1. Open two browser instances with the profiles from prerequisites.
2. Focus the first browser instance and delete any note.
3. Wait until the "Synced (hour)" message is displayed in the "Notes List" View.
4. Focus the other browser instance and click the "Sync Your Notes" button from the bottom-right part of the sidebar.
5. Observe the notes in the "Notes List" View.

**[Expected result]:**
- The same notes are displayed in the "Notes List" view on both browser instances.

**[Actual result]:**
- The deleted note on the first browser instance is still displayed in the second one.

**[Notes]:**
- The issue is reproducible also on two different devices which are synced with the same Firefox Notes profile.
- Attached a screen recording of the issue:
![deleted_notes](https://user-images.githubusercontent.com/20083658/38195520-7f0f4008-3686-11e8-9c78-5fabda4f2cdf.gif)

vladikoff:
Should be fixed in the next beta

sebastienbarbier:
Ooooh I see, I just realised this issue occured between two different Firefox profile, which is not what I was testing 😱. Sorry, misread it. Will do some test.

vladikoff:
Updating issue title to clarify

sebastienbarbier:
After many investigations, seams like we do not implement kinto.js **sync process** properly.

Our sync does not send on request the `_since` parameter _(which should be automatically managed)_ so as a result we always retrieve the entire list instead of only the differences.

On delete, server create a **tombstone** which confirm it is properly handled, but those are only returned if a `_since` parameter is provided. Tombstones when received should trigger `load event` event and update properly our UI.

I could not manage to make this work, and will only be able to continue working this on next Monday. @vladikoff I guess we need to decide if this is blocking v4 release.

vladikoff:
cc @natim

vladikoff:
@Natim === 🚒👨‍🚒🎇

glasserc:
[...]

PR #906, opened 2018-04-06
Sync on create and delete

### LINKED ISSUE #872: "Send to Notes" creates a new note but does not sync the new note
@sebastienbarbier see https://github.com/mozilla/notes/commit/c420ae8d26dc706f74d7a5dab704d666ef13a524#diff-4e21b49273243fca121200053b009515R114

I had to make the change to update the synced status in that PR but it doesn't actually send the new note to kinto. What should we use there?

sebastienbarbier:
This PR implement sync on create and delete. Will be needed for #858 but I guessed would not nice to split in a dedicated PR.
Fix #872

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -206,7 +206,7 @@ browser.runtime.onMessage.addListener(function(eventData) {
       break;
     case 'delete-note':
       // We create a note, and send id with note-created nessage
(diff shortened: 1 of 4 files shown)


sebastienbarbier:
Will be merged and review with PR #917

PR #917, opened 2018-04-09
Sync across profile

### LINKED ISSUE #858: The Notes List is not correctly synchronized across Firefox profiles when a note is deleted

### LINKED ISSUE #872: "Send to Notes" creates a new note but does not sync the new note

sebastienbarbier:
Implement proper syncing by manually tracking lastSync event.

delete

Fix #858 Fix #872


src/sidebar/app/actions.js
```diff
@@ -101,7 +101,8 @@ export function createNote(content = '', origin) {
     id,
     content,
     origin,
-    lastModified: new Date()
+    lastModified: new Date(),
```
Natim: This is a bit tedious, we shoud use the server timestamp here. I think Kinto.js is supposed to keep it :/

Natim: https://github.com/Kinto/kinto.js/blob/master/src/collection.js#L286

src/sync.js
```diff
@@ -202,6 +204,8 @@ function syncKinto(client, credentials) {
       });
     })
     .then(syncResult => {
+      lastSyncTimestamp = new Date().getTime(); // eslint-disable-line no-undef
```
Natim: This is a bit tedious, we shoud use the server timestamp here. I think Kinto.js is supposed to keep it :/
https://github.com/Kinto/kinto.js/blob/master/src/collection.js#L286

src/sidebar/app/actions.js
Natim: kinto.js is supposed to add it as soon and only as soon as the record is created remotely.

Natim:
I'm glad you were able to fix the getList to use the `_since` parameter.
However I believe we still have somehow a misuse of kinto.js that we should investigate.

src/sidebar/app/actions.js
sebastienbarbier: This is confusing I agree, but we keep inside our blob an other lastModified date for the UI which is only updated by our redux reducer. Sync event was sometimes changing dates and that was confusing for the user. Also non-synced object had no dates ... so that why. We could have done differently ? 🤔

src/sync.js
Natim: Not yet but that's something we should definitely investigate.

sebastienbarbier: That's probably why kinto.js is not handling this by itself and need our dirty patch 😓.

src/sidebar/app/actions.js
Natim: Oh yeah you are right `lastModified` and `last_modified`

sebastienbarbier: Quick vote @Natim , should I rename it lastEdited to avoid any confusion ?

Natim: lastModifiedDate ?

vladikoff:
@sebastienbarbier 1-2 tests failing:

```
FAILED TESTS:
  Authorization
    syncKinto
      ✖ should handle a conflict (slow: 1.874 secs)
        Firefox 59.0.0 (Linux 0.0.0)
      No fallback response defined for GET to https://example.com/v1/buckets/default/collections/notes/records?_sort=-last_modified&_since=1523265666782
      FetchMock.prototype.fetchMock@test/dist/unit-bundle.js:54529:10
      timedFetch/<@src/vendor/kinto.js:4064:9
      $Promise@src/vendor/kinto-http.js:596:3141
      timedFetch@src/vendor/kinto.js:4045:14
      _callee3$@src/vendor/kinto.js:4226:24
```

src/sync.js
sebastienbarbier: @vladikoff fyi, **PR is working** but this is the dirty solution I had to implement. We keep track in sync.js instead of Kinto.Collection object which return for unknown reason lastModified as `NaN`. Except that point, it's 100% handled by kinto as expected *(no extra fetch as discussed)*.

We might have an issue having some gap between real sync and `new Date()` (could we save that date before requesting so worst case we have extra redundant data from those millisecond on next sync ?).

Once again we're dirty and it shows something is wrong in our process with Kinto. I would suggest to open an extra bug and keep working on it when having more time.

sebastienbarbier:
Found a bug on deletion accorss profile. When syncing while editing a note which has been deleted by an other profile, editor keep the deleted note on display.
bugondeleteeditor

sebastienbarbier:
I applied a fix, and it's pretty clear what is going on having both window next to each other, but might be confusing if you sync and forgot about having deleted that specific note just seeing your editor going blank ... 🤔 ? @vladikoff not best UX but what do you think ?
fixedbugondeleteeditor

Natim:
IMHO it should return to the list of notes in the other editor.

PR #930, opened 2018-04-11
Using redux for profile data

### LINKED ISSUE #919: Use redux for profile data
@sebastienbarbier it's a bit sketchy right now: 

https://github.com/mozilla/notes/blob/master/native/app/components/DrawerItems.js#L61-L67

and doesn't update on change thing

Here's where we save the creds:

https://github.com/mozilla/notes/blob/master/native/app/vendor/fxa-utils.js#L72

sebastienbarbier:
Seams like profile returned by fxaUtils.fxaGetCredential() does not contain a `displayName`, but except that profile is not stored in redux 🙂

This PR store `email`, `displayName`, and `avatar` in redux `state.sync` object, might be nice to have a `state.profile` but I would like to keep same syntax as our web-ext for now and propose a refactoring to share actions and reducers between both app (suggestion to be discussed).

Fix #919


vladikoff:
> Seams like profile returned by fxaUtils.fxaGetCredential() does not contain a displayName

@sebastienbarbier displayName will only be returned if the user sets a display name at https://accounts.firefox.com/settings (and then we relogin and refresh the creds)

native/app/components/DrawerItems.js
```diff
@@ -78,10 +71,10 @@ class DrawerItems extends React.Component {
           borderColor='#FFFFFF'
           borderWidth={2}
           resizeMode='cover'
-          source={{uri: this.state.profileAvatar}}
+          source={{uri: this.props.state.sync.avatar}}
```
vladikoff: @sebastienbarbier i had issues here where if this is ever undefined (passed into `uri` then it all show a red error). I am not sure if we need to do a check earlier in the `jsx` or not

vladikoff:
if you are hanging out in this view, the new design for this is here https://mozilla.invisionapp.com/share/BTGS26C2FE4#/screens/290387768 + let's keep the white border around the avatar

vladikoff:
@sebastienbarbier could you check something, swipe the app away from the process and reload, do you get information back into the drawer:

vladikoff:
Maybe related https://github.com/react-navigation/react-navigation/issues/246 but could be something else

native/app/components/SplashPanel.js
```diff
@@ -19,7 +28,7 @@ class SplashPanel extends React.Component {
           actions: [NavigationActions.navigate({ routeName: 'LoginPanel' })],
         }));
       }
-    });
+    }).catch((e) => console.error(e));
```
vladikoff: nit: if something fails here let's try do to a navigation to the login panel. instead of just dropping the error?

sebastienbarbier:
After spending half a day trying to merge react-navigation state and our redux ([reactnavigation.org/docs/redux-integration](https://reactnavigation.org/docs/redux-integration.html)), seams like that was not needed and found some issues I could fix 😅. 
At least I learned a lot.

I think I solved it for now ... @vladikoff can you give a try and confirm ?

native/app/components/DrawerItems.js
sebastienbarbier: This should not be undefined ... I'll investigate about that but had no issue for now.

PR #978, opened 2018-04-23
Pull to refresh now update the "Last Synced" date

### LINKED ISSUE #971: Pull to refresh does not update the "Last Synced" date
time is 9:52 , but it says 9:32 in the footer

sebastienbarbier:
- [x] Update the "Last synced" date.
- [x] Display Snackbar on listPanel only, but on all sync event.
- [x] When syncing from drawer, add delay to trigger snackbar only AFTER closing the drawer

Fix #971

### DIFF
--- a/native/app/components/ListPanel.js
+++ b/native/app/components/ListPanel.js
@@ -28,23 +28,28 @@ class ListPanel extends React.Component {
         action: KINTO_LOADED
       });
     }
+
+    this._keyExtractor = (item, index) => item.id;
+  }
+
+  _triggerSnackbar = () => {
+    this.setState({
+      refreshing: false,
+      snackbarSyncedvisible: this.props.navigation.isFocused()
+    });
   }
 
   componentWillReceiveProps(newProps) {
     if (this.props.state.sync.isSyncing && !newProps.state.sync.isSyncing) {
-      this.setState({
-        refreshing: false,
-        snackbarSyncedvisible: this.props.navigation.isFocused()
-      });
+      if (this.props.state.sync.isSyncingFrom === 'drawer') {
+        setTimeout(this._triggerSnackbar, 400);
+      } else {
+        this._triggerSnackbar();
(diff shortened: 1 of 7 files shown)


native/app/components/ListPanel.js
```diff
@@ -23,13 +24,28 @@ class ListPanel extends React.Component {
...
-        this.setState({ snackbarSyncedvisible: true })
+      browser.runtime.sendMessage({
+        action: KINTO_LOADED
       });
     }
   }
 
+  componentWillReceiveProps(newProps) {
+    if (!newProps.state.sync.isSyncing) {
+      // We do not display snackbar on sync if triggered from other than pullRefresh
```
vladikoff: What happens if we change this and always show the snackbar, no matter how you got the new notes?

sebastienbarbier: It looks weird 😅, but open to any suggestion.
snackbaroverdrawer

vladikoff: > Nothing wrong, but feel to me like a bit intrusive ... what do you think ?

To clarify, we only show the Snackbar if the user is on the ListPanel view. In the single note view we have the header text to show the status i guess

sebastienbarbier: I like your suggestions and will apply right away ! 😁.

vladikoff: `elevation` might not be the right thing to stack the Snackbar behind the drawer. Glancing at the pr it's not clear to me at this time how the thing shows up above the Drawer in our case. But we might not have to worry about that because we can time it such that we close the Drawer first and then show the snackbar.

(ref: https://github.com/callstack/react-native-paper/pull/305/files )

sebastienbarbier:
Had to refactor kintoLoad to handle errors from drawer sync button. 
We should now be using promises through `dispatch(kintoLoad())` 👍

PR #979, opened 2018-04-23
Delay redux disconnect event to avoid empty UI on logout

### LINKED ISSUE #968: Don't block the UI on logout
You can see a disappearing profile info below, sometimes it also hangs if other operations such as Kinto reset and Keystore clear take longer than expected

sebastienbarbier:
You can guess that I am not super happy about how this PR has been implemented, but it works and catching drawer closing events does not seam like an easy option 😥.

logoutui

Fix #968

### DIFF
--- a/native/app/components/DrawerItems.js
+++ b/native/app/components/DrawerItems.js
@@ -48,7 +48,11 @@ class DrawerItems extends React.Component {
       {
         label : 'Log out',
         action: () => {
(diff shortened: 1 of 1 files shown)


vladikoff:
> You can guess that I am not super happy about how this PR has been implemented, but it works and catching drawer closing events does not seam like an easy option

Right, this might not work on slower devices, but let's see how it goes

PR #980, opened 2018-04-23
Load the ListPanel as soon as possible

sebastienbarbier:
- [x] Load the ListPanel as soon as possible
- [x] Add a progress bar (from paper module), while checking if there are any notes in kinto
- [ ] Check redux state, if empty, show No Notes
- [ ] Trigger sync using the "on activity resume", update the list when new notes arrive.


sebastienbarbier:
Will make conflicts with PR #978 which refactor the loading process, but should be super easy to understand how to fix 🙃.

vladikoff:
Menu button key gets highlighted on login:

PR #1012, opened 2018-04-30
Implement Snackbar notification for Notes Deleted 

### LINKED ISSUE #926: Implement Snackbar notification for Notes Deleted
For Delete and Synced

**We should be able to use this: github.com/callstack/react-native-paper/pull/305/files**

******

image

image

### LINKED ISSUE #1018: Implement our own Snackbar so we can Animate.parallel the FAB
@sebastienbarbier let's replace the paper snackbar with our own, so we have direct access to the animation and stacking order

### LINKED ISSUE #1020: The fab button should be 24(px?) from bottom and right edge
Feedback from UX, we might have to match it with the mock instead of using the `24` value directly

sebastienbarbier:
- [x] On delete, display blue snackbar (#926)
- [x] Override default snackbar to animate.parallel with FAB (#1018) 
- [x] The fab button should be 24(px?) from bottom and right edge (#1020 )
- [x] Implement UNDO button

Fix #926, fix #1018, fix #1020

### DIFF
--- a/native/app/actions.js
+++ b/native/app/actions.js
@@ -50,27 +50,34 @@ export function authenticate(loginDetails) {
   return { type: SYNC_AUTHENTICATED, loginDetails };
 }
 
-export function createNote(content = '') {
+export function createNote(note = {}) {
   // Return id to callback using promises
   return (dispatch, getState) => {
     return new Promise((resolve, reject) => {
 
-      const id = uuid4();
+      note.id = uuid4();
+      if (!note.lastModified) note.lastModified = new Date();
 
       // If note is not a paragraph, we force creation but editor should also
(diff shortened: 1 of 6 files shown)


sebastienbarbier:
**Implemented the animation** 🎉, looks awesome! but undo delete is now broken. _Should be easy to fix._
 
floatingfab
floatingfabdelete

sebastienbarbier:
Aie, bug: Sync snackbar doesn't work anymore after deleting a note, and I can't find why. Will investigate tomorrow, back with `WIP` label 😓 !

sebastienbarbier:
@vladikoff Can you do some testing, should works fine now 👍

PR #1014, opened 2018-04-30
Alternative loading screen

sebastienbarbier:
Could this be an option ?

loadingscreen


vladikoff:
@antlam thoughts?

vladikoff:
please rebase

antlam:
> Could this be an option ?

Skeleton UI is cool and there are perceived perf benefits too. 

But I think it also needs more thought.. E.g. what does the animation look like, how much of the actual UI should it imitate? and what our our motion design guidelines here (from Photon), etc...

vladikoff:
per @antlam's comment let's focus on the rest of the bugs and issues. maybe bring this back later.

PR #1015, opened 2018-04-30
Add loading animation on signin button

### LINKED ISSUE #948: Finalize the "Sign in" button design
1. Add a Loading Indicator. For example: https://github.com/callstack/react-native-paper/blob/master/example/src/ButtonExample.js#L55 On some devices it takes a bit to open up the login browser.

image

2. Use the design from the mock up:

image

sebastienbarbier:
On click, signin button now show a loading animation. 
Current action also redirect to `LoadingPanel` which hide the animation ... not sure if we want to keep this behaviour or if the animation still make sense.

singinanimation

Fix #948

### DIFF
--- a/native/app/components/LoginPanel.js
+++ b/native/app/components/LoginPanel.js
@@ -14,7 +14,16 @@ import { KINTO_LOADED } from '../utils/constants';
 import browser from '../browser';
 
 class LoginPanel extends React.Component {
+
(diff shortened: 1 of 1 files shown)


vladikoff:
Can we do anything about that?

kapture 2018-04-30 at 10 43 33

PR #1021, opened 2018-05-02
The fab button should be 24(px?) from bottom and right edge

### LINKED ISSUE #1020: The fab button should be 24(px?) from bottom and right edge

sebastienbarbier:
capture d ecran 2018-05-02 a 10 40 51

Fix #1020


sebastienbarbier:
Will be fixed in PR #1012

PR #1053, opened 2018-05-11
Multi delete feature

### LINKED ISSUE #1038: [Android] The "Undo" option is not displayed if a note is deleted while the "Notes Synced!" banner is displayed
**[Affected Platforms]:**
- Samsung Galaxy S8 (Android 8.0.0)
- Pixel (Android 8.0.0)
- Huawei P8 Lite (Android 6.0)

**[Prerequisites]:**
- Have the latest version of the "Firefox Notes" application (notes-1_0-qa-v2046 custom built on 07.05.2018) installed.
- Have at least a note in the application.

**[Steps to reproduce]:**
1. Open the Notes application and log in if needed.
2. Swipe down to refresh the notes and tap on any note.
3. Tap the menu button from the top right part of the page.
4. Select the "Delete" option while the "Notes Synced!" message is still displayed.
5. Observe the behavior.

**[Expected result]:**
- The note is deleted, you are returned to the list view and the "Undo" banner is displayed.

**[Actual result]:**
- The note is deleted, you are returned to the list view, but the "Undo" banner is not displayed.

**[Notes]:**
- The issue is reproducible only if the note is deleted while the "Notes Synced!" message is displayed.
rec of the issue

sebastienbarbier:
Implement multi delete, and stack notification to avoid skipping one (fix #1038).

multidelete

This PR also remove `github:vladikoff/react-native-paper#notes` from package.js 👍


native/app/components/ListPanelHeader.js
```diff
@@ -10,15 +12,56 @@ class ListPanelHeader extends Component {
...
+          <ToolbarAction
+            size={20}
+            style={{ paddingTop: 4 }}
+            color='white'
+            icon='clear'
+            onPress={() => this._resetSelection()} />
+          <ToolbarContent
+            style={{ paddingLeft: 0,  }}
+            titleStyle={{ fontSize: 18, color: 'white' }}
+            title={ `${this.props.state.sync.selected.length} selected`} />
```
vladikoff: This needs to be properly translated.

native/app/components/ListPanelHeader.js
```diff
@@ -10,15 +12,56 @@ class ListPanelHeader extends Component {
...
+      const state = this.props.state;
+      const notes = state.notes.filter((note) => state.sync.selected.includes(note.id));
+
+      if (state.sync.selected) {
+        this.props.dispatch(deleteNotes(state.sync.selected, 'multi-delete'));
+      }
+      this.props.navigation.navigate('ListPanel', { deletedNote: notes });
+
+      // console.log(notes);
+      // props.dispatch(resetSelect());
```
vladikoff: Remove this

native/app/components/ListPanelHeader.js
```diff
@@ -10,15 +12,56 @@ class ListPanelHeader extends Component {
...
+        </Toolbar>
+      );
+    }
+
     return (
       <Toolbar style={styles.toolbar}>
         <ToolbarAction
           size={20}
           style={{ paddingTop: 4 }}
+          color='#4173CE'
```
vladikoff: should this use a constant color setting?

native/app/components/ListPanelHeader.js
```diff
@@ -10,15 +12,56 @@ class ListPanelHeader extends Component {
...
 
   render() {
     const { navigation } = this.props;
+    if (this.props.state.sync.selected) {
+      return (
+        <Toolbar style={styles.toolbarSelected}>
+          <ToolbarAction
+            size={20}
+            style={{ paddingTop: 4 }}
+            color='white'
```
vladikoff: use color from constants?

native/app/actions.js
```diff
@@ -100,16 +101,16 @@ export function updateNote(id, content, lastModified) {
   return { type: UPDATE_NOTE, id, content, lastModified };
 }
 
-export function deleteNote(id, origin) {
-  trackEvent('new-note', {
+export function deleteNotes(ids = [], origin) {
+  trackEvent('delete-note', {
```
vladikoff: is there an inconsistency introduced here? does the web extension send this event for 1 note, and the mobile app will send a single event even if multiple notes deleted?

native/app/components/ListItem.js
```diff
@@ -80,6 +107,9 @@ const styles = StyleSheet.create({
     justifyContent: 'space-between',
     alignItems: 'center'
   },
+  selected: {
+    backgroundColor: 'rgb(236, 241, 250)'
```
vladikoff: move this color to constants as well

native/app/components/ListPanel.js
```diff
@@ -132,7 +140,7 @@ class ListPanel extends React.Component {
 
         // Show snackbar
         this._showSnackbar({
-          text: 'Note deleted!',
+          text: `Deleted ${ deletedNote.length } note${ deletedNote.length > 1 ? 's' : '' }`,
```
vladikoff: This needs to be properly translated

vladikoff: @flodolo How would we do this (in web extensions for example)

flodolo: in Webextensions there's no way to do it, because WE don't support plural (meh…).

So, you would need to do something like "Delete notes ($number)".

In Fluent, you could do it with a proper plural
https://projectfluent.org/fluent/guide/selectors.html

native/app/components/ListPanelHeader.js
flodolo: Sadly, not really, adjectives are declined based on the number and need a proper plural form. For example, in Italian:
* 1 selected = 1 selezionato
* 2 selected = 2 selezionati

vladikoff: @flodolo ok thank you!!

@sebastienbarbier we need a string that doesn't have numbers here

flodolo: Side question: are you going to use message.json to localize the add-on? It's not supported directly by Pontoon, so you'd need to go through properties -> JSON conversion.

Not sure if I'm looking at a mobile app or something else

vladikoff:
@sebastienbarbier needs rebase, let's fix the English strings so they don't have numbers in them and clean up. Then we can convert the strings to use the translation API as part of the `i18n` pull request

native/app/actions.js
sebastienbarbier: Great point 🤔. It should definitely send one request per note deleted. 👍

PR #1108, opened 2018-05-22
Add keystroke for Ctrl/Cmd+S to ignore Save As popup

### LINKED ISSUE #955: prevent Save As default action when I cmd + s inside an active Note
### [Description]:

When I hit cmd/ctrl + s inside Notes, I don't want to save the adjacent webpage.

### [Affected versions]:
Latest.

### [Steps to reproduce]:
1. open Notes
2. inside a note, hit cmd/ctrl +s

### [Expected result]:

Nothing, or a Sync.

### [Actual result]:
Save As dialog (for the tab that just happens to be open).

### [Additional Notes]:
I'm trying to move over from Evernote, and cmd+s is a habit of mine -- it forces a sync there.

Also, Notes is the greatest. Thank u for ur service.

vladikoff:
Fixes #955.


vladikoff:
Thanks @cedricium !!

PR #1161, opened 2018-06-14
Since some updates happend, here are the changes mandatory to fix that.

Natim:
r? @vladikoff

### DIFF
--- a/src/sync.js
+++ b/src/sync.js
@@ -173,7 +173,7 @@ function syncKinto(client, credentials) {
       credential = received;
 
       if (!received) {
-        return Promise.resolve();
+        return Promise.resolve({});
       }
 
       return fxaRenewCredential(credential)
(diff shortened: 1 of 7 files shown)


src/sync.js
```diff
@@ -172,7 +172,9 @@ function syncKinto(client, credentials) {
     .then(received => {
       credential = received;
 
-      if (!received) return;
+      if (!received) {
+        return Promise.resolve();
```
Natim: return null

src/sync.js
```diff
@@ -254,6 +256,7 @@ function syncKinto(client, credentials) {
       } else if (syncResult === undefined) {
         throw new Error('syncResult is undefined.');
       }
+      return Promise.resolve();
```
Natim: return null

Issue #1186, opened 2018-06-22
The spacing between paragraphs is not kept in the browser notes if a text was pasted in a note from the "Notes" mobile application


Softvision-CristinaBadescu:
**[Affected Platforms]:**
- Android 8.0.0 (Google Pixel)
- Android 8.0.0 (Samsung S8)
- Android 6.0 (Huawei P8 Lite)

**[Prerequisites]:**
- Have the latest version of the "Firefox Notes" application (notes-1_0-qa-v2393 custom build created on 22.06.2018) installed.
- Have a new Firefox profile with the latest version of "Firefox Notes" add-on (4.2.0dev custom build created on 22.06.2018) installed.
- You are logged with the same account in both "Notes" instances.
- On mobile, have a text with multiple paragraphs copied to the clipboard from other app.

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Open the "Notes" mobile application and log in if needed.
3. Open a new note.
4. Paste the text from prerequisites and observe the behavior.
5. Go to the browser instance of "Notes" and click the "Sync Your Notes" button.
6. Click the newly created note and observe the behavior. 

**[Expected result]:**
Step 4 & 6 - The note is displayed and the spacing is kept.

**[Actual result]:**
Step 4 - The note is displayed and the spacing is kept.
Step 6 - The note is displayed, but the spacing is not kept.

**[Notes]:**
- Here is a screen recording of the issue:
![notes - no spacing on desktop instance](https://user-images.githubusercontent.com/31244235/41781927-650c2b44-7641-11e8-8896-772eabcae3f5.gif)

Softvision-CristinaBadescu:
This issue is no longer reproducible using the latest version of the "Firefox Notes" application (notes-1_0-qa-v2516 custom built on 25.06.2018). Verified on Samsung S8 (Android 8.0.0), Huawei P8 Lite (Android 6.0) and Pixel (Android 8.0.0).

Issue #628, opened 2018-01-30
Content copied from "Notepad++" or "Skype" does not keep its format after it's pasted in "Firefox Notes"


Softvision-MariusComan:
**[Affected versions]:**
- Firefox 58.0 and up

**[Affected Platforms]:**
- All Windows
- All Mac
- All Linux

**[Prerequisites]:**
- Have a Firefox profile with the latest version of the "Firefox Notes" add-on (2.2.0rc2) installed.
- Have content on multiple rows, copied to clipboard from "Notepad++" or "Skype".

**[Steps to reproduce]:**
1. Open the browser with the profile from prerequisites.
2. Paste the content from prerequisites.
3. Observe the pasted content.

**[Expected result]:**
- The pasted content is displayed on multiple rows.

**[Actual result]:**
- The pasted content is displayed on the same row.

**[Additional Notes]:**
- The issue is reproducible with the TxP version of "Firefox Notes".
- Attached a screen recording of the issue:
![rec of the issue](https://user-images.githubusercontent.com/31243259/35560812-2b4f84d2-05b7-11e8-962d-14b304b82b4b.gif)

Softvision-MariusComan:
This issue is no longer reproducible with the latest version of the "Firefox Notes" add-on (4.2.0 RC2) installed, on the latest Firefox release and Nightly. Verified on Windows 10 x64, Mac 10.13 and Ubuntu 14.04 x64.

Issue #534, opened 2017-12-28
Issue with Line Break on Mac OS


diogocnunes:
I haven't tested this on Windows. Therefore, I will assume that the error only occurs in MacOS.

I use Firefox Notes on Mozilla Firefox 57.0.3 on MacOS 10.13.2 High Sierra. When copying a text from Text Editor or Sublime Text to Firefox Notes, line breaks are removed and the text is all on the same line ...

If I correct the text in Firefox Notes and copy it to the Text Editor or Sublime Text, in the applications the line breaks are duplicated ...

Copying and pasting between editors, the problem doesn't occur.

I suggest putting an option to configure the default line break (Unix, Mac, Windows, ...)

diogocnunes:
Examples:

**How shows in Sublime:**

![captura de tela 2017-12-28 as 23 50 44](https://user-images.githubusercontent.com/13525900/34426009-0320ea56-ec2a-11e7-9f90-330824711fb4.png)

**How shows in Firefox Notes:**

![captura de tela 2017-12-28 as 23 50 59](https://user-images.githubusercontent.com/13525900/34426013-0b27f83e-ec2a-11e7-9ee6-49c40eb7b934.png)

Natim:
Thank you for posting this issue, it is indeed very inconvenient.

@vladikoff I guess this is because linebreaks should be converted to <br/> on Text Plain paste?

vladikoff:
cc @mgsy

Mgsy:
Thanks for the information. This problem occurs not only on OS X, but also on Windows. It's caused by way how editor converts plaint text to HTML. Current conversion algorithm looks like this:
```
“a\nb” -> “a<br>b”
“a\n\nb” -> “<p>a</p><p>b</p>”
```
As we don't handle soft-line breaks, you receive the output from this issue. I've opened a discussion ticket in our CKEditor 5 repository - https://github.com/ckeditor/ckeditor5/issues/766. You can monitor it and feel free to join the discussion.

PR #1191, opened 2018-06-23
Update to latest CKEditor

vladikoff:
Fixes https://github.com/mozilla/notes/issues/1186
Fixes https://github.com/mozilla/notes/issues/534
Fixes https://github.com/mozilla/notes/issues/628


ckeditor-build/src/ckeditor.js
```diff
@@ -1,46 +1,51 @@
...
-      'code'
-    ]
+    toolbar: {
+      items: [
+        'headings',
+        'bulletedList',
+        'numberedList',
+        'bold',
+        'italic',
+        'strike',
```
pomek: It won't work. The correct name is `strikethrough`. See https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/features/basic-styles.html.

vladikoff: So we have the config here: https://github.com/mozilla/notes/blob/master/src/sidebar/app/data/editorConfig.js#L14 

I always get confused why we need this array here anyway, because we initialize with a different config.  Are these the defaults or something?

pomek:
This is default config for the build. If you won't specify the toolbar items during the editor initialization (`ClassicEditor.create( element )`), the editor will use the built-in config.

I guess, in your case the `ClassicEditor.build` property could be skipped.

ckeditor-build/src/ckeditor.js
pomek: This is default config for the build. If you won't specify the toolbar items during the editor initialization (`ClassicEditor.create( element )`), the editor will use the built-in config.

I guess, in your case the `ClassicEditor.build` property could be skipped.

PR #1267, opened 2018-07-05
Add logic to sign the APK from circleci

jvehent:
This should automate apk signing for everything pushed to the master branch. You can also require a tag by tweaking that tags.only regex.
I set the env var in circleci already, so it should just work :tm:


vladikoff:
@jvehent somehow this didn't even trigger Circle CI building. Syntax error or something?

PR #1327, opened 2018-07-30
Split APK into x86 and armv7 flavors

carolkng:
APK size without split: 13M
APK w/ split (x86): 9.7M, (armv7): 8.3M

r? @vladikoff


native/android/app/build.gradle
```diff
@@ -128,6 +129,14 @@ android {
             include "armeabi-v7a", "x86"
         }
     }
+    productFlavors {
```
ncalexan: There's a whole different way to achieve this in the Android-Gradle plugin: https://developer.android.com/studio/build/configure-apk-splits#configure-abi-split.  Explain to me what the differences between this approach and that approach are?

PR #1328, opened 2018-07-30
Refactor token renewal to use fxa client

carolkng:
Also removes the need to parse the JSON account object

r? @vladikoff @eoger

### DIFF
--- a/native/android/app/src/main/java/com/notes/fxaclient/FxaClientModule.java
+++ b/native/android/app/src/main/java/com/notes/fxaclient/FxaClientModule.java
@@ -3,7 +3,6 @@ package com.notes.fxaclient;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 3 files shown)


native/android/app/src/main/java/com/notes/fxaclient/FxaClientModule.java
```diff
@@ -96,19 +112,55 @@ public void begin(final Callback successCallback,
...
     }
 
+    @ReactMethod
+    public void renewToken(String state, final Callback successCallback,
+                           final Callback errorCallback) {
+        FxaResult<FirefoxAccount> account;
+        if (this.account != null) {
+            account = FxaResult.Companion.fromValue(this.account);
+        } else if (state != null) {
+            Log.e("called renew", "help");
```
vladikoff: We can do proper logging here if we want to . Maybe added proper logging to other branches as well?

PR #1341, opened 2018-08-04
[WIP] Support export to email-tabs

ianb:
See also https://github.com/mozilla/email-tabs/pull/108

This doesn't implement registration properly, but has some stubby code to that effect (as well as hardcoded support for email-tabs).

This PR is kind of a discussion starting point.


ianb:
Doesn't seem like there's any interest

PR #1364, opened 2018-09-18
Added 3 new integration tests.

jrbenny35:
1. ```should be able to delete a note```: tests that a user is able to delete a note.
2. ```should be able to add a note from the note page```: tests that you can create a new note from the note page.
3. ```should be navigate to feedback page```: tests that the 'Give Feedback' button routes to the correct page.

### DIFF
--- a/test/integration/page_objects/note_page.js
+++ b/test/integration/page_objects/note_page.js
@@ -94,16 +94,18 @@ export default class NotePage extends BasePage {
   * @returns {Object} An object representing the page.
   * @throws ElementNotFound
   */
-  async addNote(title = "This is a title", paragraph = 'This is a test!') {
+  async addNote(title = 'This is a title', paragraph = 'This is a test!') {
     this.logger.info('Adding a new note');
     let editor = await this.driver.findElement(this.by.css(this.editorLocator));
-    await editor.sendKeys(title, this.key.ENTER, paragraph);
+    await editor.sendKeys(title);
+    await editor.sendKeys(this.key.ENTER);
+    await editor.sendKeys(paragraph);
     // Find note title before we wait for it to change
     let thisNoteTitle = await this
       .driver
       .findElement(this.by.css(this.noteTitleLocator));
     await this.driver.wait(
-        this.until.elementTextIs(thisNoteTitle, paragraph),
+        this.until.elementTextIs(thisNoteTitle, title),
       this.timeout
     );
     this.logger.info('Note added');
--- a/test/integration/page_objects/note_page.js
+++ b/test/integration/page_objects/note_page.js
@@ -7,7 +7,14 @@ export default class NotePage extends BasePage {
     super(driver);
     this.backBtnLocator = '.iconBtn';
     this.editorLocator = '.ck-editor__editable';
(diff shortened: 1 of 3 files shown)


test/integration/testFirefoxWebext.js
```diff
@@ -91,4 +91,39 @@ describe('The Firefox Notes web extension', function() {
...
+  });
+
+  it('should be able to add a note from the note page', async function() {
+    // Add a note from the Note Page
+    let listPage = await new ListPage(driver).waitForPageToLoad();
+    let notePage = await listPage.newNoteButton();
+    let title = "THIS IS A TEST";
+    let paragraph = "this isnt a test";
+    await notePage.addNote(title, paragraph)
+    expect(await notePage.noteTitle).to.equal(paragraph)
```
pdehaan: A few more missing semicolons here, but maybe we're not running ESLint against the test directories.

This also had me a bit confused as it looks like we're asserting that the `noteTitle` is equal to the `paragraph` (I kind of expected it to equal the `title`, but I don't know the underlying data/objects).

pdehaan:
Overall looks great. I had a bit of confusion around why ESLint isn't enforcing rules in these files, but that's outside the scope of this PR.

Also confused why we're adding an updated package-lock.json file when the dependencies don't seem to have changed.

R+=me

test/integration/page_objects/note_page.js
```diff
@@ -42,14 +49,52 @@ export default class NotePage extends BasePage {
...
+
   /**
   * @function addNote
   * @param  {string} title Title of the note you want top add
   * @param  {string} paragraph Paragraph of the note you want to add
   * @returns {Object} An object representing the page.
   * @throws ElementNotFound
   */
-  async addNote(title, paragraph) {
+  async addNote(title = "This is a title", paragraph = 'This is a test!') {
```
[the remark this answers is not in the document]
jrbenny35: I'll fix these.

test/integration/testFirefoxWebext.js
```diff
@@ -91,4 +91,39 @@ describe('The Firefox Notes web extension', function() {
     expect(listNoteTitle).to.equal(paragraph);
   });
 
+  it('should be able to delete a note', async function() {
+    // Deletes a note
+    let listPage = await new ListPage(driver).waitForPageToLoad();
+    let notePage = await listPage.newNoteButton();
+    let title = "THIS IS A TEST";
+    await notePage.addNote(title)
```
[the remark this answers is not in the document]
jrbenny35: So I figured that is what the original delete test was doing so I didn't need to do that again.

test/integration/testFirefoxWebext.js
jrbenny35: I think I need to change the nomenclature. The title and paragraph doesn't make much sense.

jrbenny35: @cedricium that helps a bunch actually. Thanks!

jrbenny35: Looks like I just had the 'enter' command in the wrong place.

vladikoff:
Thanks!

PR #1367, opened 2018-10-19
Change mocha command for integration tests to use babel compiler.

jrbenny35:
This gets rid of the babel build step and uses Mocha's built in feature for compiling ES6 modules.


pdehaan:
I think the package-lock.json edit is causing the CI to break.
Other than that, R+ when the CI is green.

PR #1431, opened 2019-01-07
Bug 1511391 - Fix the extension on Nightly and Beta.

emilio:
For some reason the UA sheet loaded for extensions has a -moz-user-select: none
rule.

I have absolutely no clue why it's there, the commit message that introduced it
doesn't explain much.

In any case this is a very low-risk fix for the extension.


vladikoff:
Thank you!!!

PR #1509, opened 2019-11-26
Upgrade to React-Native 60

Ilphrin:
This PR is an attempt to upgrade to RN 60. A few things to consider:

- Still a **work in progress**, but open to discussion of course!
- I do not have a Mac, so I can't upgrade files for iOS :/ so if someone has the possibility to do it please do so! :3


vladikoff:
> I do not have a Mac, so I can't upgrade files for iOS :/ so if someone has the possibility to do it please do so! :3

We don't build for iOS, so don't worry about that

Ilphrin:
> We don't build for iOS, so don't worry about that

Oh cool, that will be easier to do then 👌

vladikoff:
> 0.00s$ sh -e /etc/init.d/xvfb start
sh: 0: Can't open /etc/init.d/xvfb
The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .


There some travis failures, this is for the web extension
We can do the following:

1. disable those tests
2. fix them on travis
3. fix them to run on circle ci

Those tests can be a bit flaky because they run in a browser.

Ilphrin:
@vladikoff I have a question (sorry for being away, christmas and all took me some of my spare time xD)
It seems like the compilation on my local machine is not working because of :
```
notes/native/android/app/src/main/java/com/notes/advancedwebview/AdvancedWebviewManager.java:9: error: package com.facebook.react.views.webview does not exist
import com.facebook.react.views.webview.ReactWebViewManager;
```
So I searched for AdvancedWebview, and it seems like it is a NativeModule that is not used in the code, do you think it is safe to remove it?

vladikoff:
> So I searched for AdvancedWebview, and it seems like it is a NativeModule that is not used in the code, do you think it is safe to remove it?

I think it would be safe to remove if it is not used !

vladikoff:
Though we should be using the following class: https://github.com/mozilla/notes/blob/623696b6c4f04b4f788e375b939cc324f8ab94c8/native/android/app/src/main/java/com/notes/advancedwebview/AdvancedWebviewManager.java#L81 to focus the keyboard when new notes are opened, so that is what it was for

Ilphrin:
> Though we should be using the following class:
> 
> https://github.com/mozilla/notes/blob/623696b6c4f04b4f788e375b939cc324f8ab94c8/native/android/app/src/main/java/com/notes/advancedwebview/AdvancedWebviewManager.java#L81
> to focus the keyboard when new notes are opened, so that is what it was for

Oh ok, so this part of the code is used? I see no places outside this folder where this class is imported 🤔

vladikoff:
> Oh ok, so this part of the code is used? I see no places outside this folder where this class is imported 🤔

Maybe it got refactored out at some point , hmm

vladikoff:
@Ilphrin I Think it's loaded here https://github.com/mozilla/notes/blob/master/native/android/app/src/main/java/com/notes/MainApplication.java#L42 ?

Ilphrin:
> @Ilphrin I Think it's loaded here https://github.com/mozilla/notes/blob/master/native/android/app/src/main/java/com/notes/MainApplication.java#L42 ?

This line would also be removed. Thanks to this line, the package is linked to the JS code and can be imported from the JS code. But I can't find any place where the app is using this package from the JS side.

Ilphrin:
Little followup, I am trying to resolve an issue where the app won't start. I have a white screen, and before anything initialize the app crashes and I can't have any log (even with `adb logcat`). As I spent a lot of time on it I think it may be safer to put this upgrade aside for now, and I may focus on other things to contribute like github issues or tinier cleanup ^^ But there is too much things I don't understand yet on the code to be able to achieve this PR I'm afraid, even though I thought I was close to finish it, sorry for that :/

vladikoff:
The crash i am seeing with this is the following: 

`java.lang.UnsatisfiedLinkError: couldn't find DSO to load: libhermes.so`

Maybe see https://stackoverflow.com/questions/57036317/react-native-java-lang-unsatisfiedlinkerror-couldnt-find-dso-to-load-libherm

vladikoff:
@Ilphrin I made some progress on updating here: https://github.com/mozilla/notes/pull/1515/commits/0f99223d670c947023cb7e7a55eb543bb8c83239

For now i am getting stuck on the bundle server not running error...

